PR #235, opened 2026-04-11
feat: legion forget <id> to permanently delete a reflection

ssilvius:
## Summary

Adds `legion forget --id <reflection-id> [--repo <safety>]` to permanently
remove a reflection from both the SQLite `reflections` table and the tantivy
search index in a single command. Destructive. No soft-delete, no undo.

## Motivation

Reflections age. A reflection written as "workaround for bug X" is valid only
until bug X is fixed. When bug X ships a fix, the workaround reflection becomes
actively harmful -- it directs the agent to keep doing the workaround after it
is no longer needed. The agent has no way to know the bug was fixed unless a
later reflection explicitly retires the old one.

Concrete example: the shingle agent carried reflection
`019d573e-e508-79d2-84ed-fdbcf5545cf8` ("RAFTERS DARK MODE TOKEN GAP") which
authorized two workarounds for a persistence bug in the rafters MCP. The
reflection was written during a real bug, confirmed by the rafters agent, and
was correct at the time. Rafters shipped v0.0.39 which fixed the bug. The
reflection was never retired. Shingle has a later reflection that says
"v0.0.39 fixes the persistence bug," but BM25 returned the stale one first
for color-related queries because it had more topical surface. Shingle ended
up inside the rafters repo trying to edit `DEFAULT_SEMANTIC_COLOR_MAPPINGS` in
`defaults.ts` -- the exact file the stale reflection named -- because the
outdated doctrine was winning the recall fight.

Before this PR there was no way to delete a reflection via the legion CLI. The
`no-direct-db` hook also blocks agents from running SQL directly. The only
options were (a) post a superseding reflection and hope BM25 outranks the old
one, or (b) edit the database outside legion's control. Both are bad. The
correct answer is "retire the bomb surgically."

`legion forget --id 019d573e-...` removes the reflection cleanly. Used live to
retire the shingle time bomb in one command.

### CLI
```
legion forget --id <reflection-id> [--repo <safety-check>]
```

- `--id`: reflection id to delete. Returns `ReflectionNotFound` if the id is
  not present.
- `--repo`: optional safety check. When set, the delete is refused unless the
  reflection's actual repo matches. Prevents nuking the wrong reflection when
  working with similarly-shaped ids.
- Prints a confirmation line on success with the repo and the first 80 chars
  of the deleted text so the operator can verify what was removed.
- Writes an audit entry with `action="delete-reflection"` so deletions are
  traceable in `legion audit`.

### Database
```rust
pub fn delete_reflection(&self, id: &str) -> Result<Reflection>
```

Returns the deleted row so the caller can confirm what was removed. Returns
`LegionError::ReflectionNotFound` if the id does not match. SQLite-only.
Callers MUST also call `SearchIndex::delete(id)` or the tantivy index will
return the deleted reflection as a ghost on the next BM25 query.

### SearchIndex
```rust
pub fn delete(&self, id: &str) -> Result<()>
```

Uses `Term::from_field_text(self.id_field, id)` + tantivy `delete_term` to
remove the matching document. Commits immediately so the next recall does not
return the ghost. No error if the id is not present -- `delete_term` is a
no-op on empty matches. The database-layer check is the authoritative
existence source.

### Error type
```rust
LegionError::ReflectionNotFound(String)
```

New thiserror variant paralleling `CardNotFound`.

## Test plan

- [x] `delete_reflection_removes_row_and_returns_deleted` (src/db.rs) -- insert,
  verify visible, delete, assert returned row matches, verify gone,
  double-delete returns `ReflectionNotFound`
- [x] `delete_removes_document_from_index` (src/search.rs) -- add two docs,
  delete one, confirm the remaining doc still surfaces and the deleted one is
  gone from BM25 results (no ghost)
- [x] `delete_nonexistent_id_is_noop` (src/search.rs) -- delete a term that
  never existed must not error; existing docs unaffected
- [x] `forget_rejects_wrong_repo_safety_check` (tests/integration.rs) --
  reflect on "kelex", call forget with `--repo rafters`, assert exit nonzero
  with "repo safety check failed" in stderr naming both repos, verify the
  reflection is still recallable after the rejected delete, then verify a
  forget with the correct `--repo` succeeds
- [x] 401 bin tests + 74 integration tests + 9 others passing
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt -- --check` clean
- [x] legion-simplify gate: clean

## Out of scope

- Sweep / batch deletion (`legion forget --all-matching "workaround"`). One-at-
  a-time delete with a safety check is the right first step; batch deletion
  needs a preview-first design to avoid accidents.
- Automatic workaround-reflection retirement tied to GitHub issue close state.
  Better long-term fix but requires cross-agent issue-close coordination.
  Separate issue territory.
- Boundary enforcement. The incident that motivated this was only possible
  because shingle's tools let it navigate outside its workdir. Deleting the
  stale reflection retires the time bomb but does not prevent a future stale
  reflection from producing the same outcome. That fix is at the hook layer,
  tracked separately.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -750,6 +750,41 @@ impl Database {
         }
     }
 
+    /// Permanently remove a reflection from the database by id.
+    ///
+    /// Returns the deleted reflection so the caller can confirm what was
+    /// removed (repo, audience, first 80 chars of text, etc). Returns
+    /// `LegionError::ReflectionNotFound` if the id does not match any
+    /// row.
+    ///
+    /// This is a HARD delete from SQLite only. Callers must also call
+    /// `SearchIndex::delete(id)` to remove the matching document from
+    /// the tantivy index, or subsequent BM25 queries will still return
+    /// the deleted reflection as a "ghost" until the next reindex.
+    ///
+    /// Destructive. No soft-delete, no undo. Used by `legion forget` to
+    /// retire stale workaround reflections, bad reflections, or personal
+    /// data that should not persist in the corpus.
+    pub fn delete_reflection(&self, id: &str) -> Result<Reflection> {
+        // Fetch first so we can return it and so a missing id produces a
+        // clear error rather than a silent zero-row delete.
+        let reflection = self
+            .get_reflection_by_id(id)?
+            .ok_or_else(|| LegionError::ReflectionNotFound(id.to_string()))?;
+
+        let rows = self.conn.execute(
+            "DELETE FROM reflections WHERE id = ?1",
+            rusqlite::params![id],
+        )?;
+        if rows == 0 {
+            // Race: reflection existed at the fetch above but was
+            // deleted before our delete ran. Surface as NotFound rather
+            // than success.
+            return Err(LegionError::ReflectionNotFound(id.to_string()));
+        }
+        Ok(reflection)
+    }
+
     /// Retrieve reflections by a list of IDs. Returns them in the order found
     /// (not necessarily the input order). Missing IDs are silently skipped.
     pub fn get_reflections_by_ids(&self, ids: &[&str]) -> Result<Vec<Reflection>> {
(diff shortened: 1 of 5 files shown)


ssilvius:
Review fixes pushed in 7ba25ca.

Addressed all three high-confidence findings from the parallel review (code-reviewer + silent-failure-hunter):

1. **Wrong error type for repo safety check** -- added `LegionError::ReflectionRepoMismatch { id, actual, expected }` as a dedicated structured variant. The `--repo` gate now returns it instead of the semantically-wrong `LegionError::WorkSource`. Display message is preserved verbatim.

2. **Rejected forget attempts left no audit trail** -- the rejection path now writes `audit(outcome="rejected", details="expected=<repo> actual=<repo>")` BEFORE returning. On a destructive command, rejected attempts are forensically relevant.

3. **Silent partial failure between SQLite and tantivy** -- restructured the delete flow so:
   - `audit(outcome="success")` is written immediately after the SQLite delete succeeds, BEFORE the index delete, so the hard delete is always recorded
   - if `index.delete` fails, a loud stderr WARNING names the orphaned id, describes the current state, and tells the operator to run `legion reindex`
   - a second `audit(outcome="partial", details="index-orphan: ...")` row is written on the orphan path
   - the underlying `Search` error still propagates via `?` so exit code is nonzero

Also flattened the three resulting `audit(&db::AuditInput { ... })` call sites into a single `write_audit(agent, outcome, details)` closure (all three share action/target_type/target_ref/task_id/source_type).

Test updates: `forget_rejects_wrong_repo_safety_check` now also calls `legion audit --action delete-reflection --json` twice to verify both the rejected and success audit entries land in the log.

401 bin + 74 integration + 9 others passing. clippy -D warnings clean, fmt clean, legion-simplify gate clean.

PR #241, opened 2026-04-14
feat(#240): legion watch add/remove/list subcommands

### LINKED ISSUE #240: feat: legion watch add/remove/list subcommands for managing watch.toml
## Context

The hero copy on runlegion.dev shows:

\`\`\`
legion watch add ~/projects/myapp
legion watch add ~/projects/api --agent backend
\`\`\`

These commands do not currently exist. Bare \`legion watch\` starts the auto-wake daemon and reads \`watch.toml\` from the data dir, but the only way to register a repo is to edit the TOML file by hand. This breaks the homepage promise that \`legion watch\` = running from zero.

A \`legion watch add\` subcommand existed in the old Bun server implementation and was dropped during the Rust port. We need it back.

## Scope

Three subcommands under \`legion watch\`, plus a minor config schema extension. Bare \`legion watch\` keeps its current daemon behavior.

### \`legion watch add <path> [--agent <name>]\`

- Positional \`<path>\`, resolved to absolute.
- Error if path does not exist or is not a directory.
- Derive project/repo name from basename of the resolved path.
- \`--agent <name>\` optional: sets a separate \`agent\` field on the repo entry. Default: agent equals the project name (i.e., the basename). Handles cases like \`platform\` agent managing the \`ledger\` repo.
- If \`watch.toml\` does not exist, create it with default values for \`poll_interval_secs\`, \`cooldown_secs\`, \`stagger_secs\`.
- Append a new \`[[repos]]\` entry. If the same path is already listed, print a message and exit 0 without modifying the file.
- Print: \`Added <name> (agent: <agent>) -> <abs-path>\`.

### \`legion watch remove <path-or-name>\`

- Accepts either an absolute path or a project/repo name.
- If multiple entries match, print an error listing the matches and exit nonzero.
- If no entries match, print a message and exit 0.
- On success, print: \`Removed <name> -> <abs-path>\`.

### \`legion watch list\`

- Print all \`[[repos]]\` entries in \`watch.toml\` in a readable format: name, agent (if different from name), path.
- If \`watch.toml\` does not exist, print: \`No watch.toml configured. Run 'legion watch add <path>' to add a repo.\` and exit 0.

### Config schema

\`watch.toml\` gains an optional \`agent\` field on each \`[[repos]]\` entry:

\`\`\`toml
[[repos]]
name = \"ledger\"
workdir = \"/Volumes/store/projects/rafters-studio/ledger\"
agent = \"platform\"   # optional; defaults to name if omitted
\`\`\`

### Daemon routing

The watch daemon currently routes signals to repos by \`name\`. Update the routing logic to prefer the \`agent\` field when present, falling back to \`name\`. This makes the \`--agent\` flag semantically meaningful.

## Acceptance Criteria

- [ ] \`legion watch add <path>\` appends a repo to \`watch.toml\` and creates the file if missing.
- [ ] \`legion watch add <path> --agent <name>\` writes both \`name\` (basename) and \`agent\` fields.
- [ ] \`legion watch add <path>\` errors cleanly if path does not exist or is not a directory.
- [ ] \`legion watch add <path>\` is idempotent: adding the same path twice does not duplicate the entry.
- [ ] \`legion watch remove <path-or-name>\` removes a matching entry and handles ambiguous / missing matches correctly.
- [ ] \`legion watch list\` prints all repos, indicating when \`agent\` differs from \`name\`.
- [ ] Config deserialization is backwards-compatible: existing \`watch.toml\` files without \`agent\` fields load without error.
- [ ] Daemon wake-routing prefers \`agent\` field when present, falls back to \`name\`.
- [ ] Integration tests cover: add creates file, add is idempotent, remove succeeds, remove on missing is graceful, list with and without config, agent field roundtrip through TOML.
- [ ] \`cargo test\`, \`cargo clippy -- -D warnings\`, \`cargo fmt -- --check\` all pass.

## Out of Scope

- No \`legion watch edit <name>\` subcommand. Users can \`remove\` + \`add\` to modify.
- No multi-agent watching per repo (one agent per \`[[repos]]\` entry).
- No changes to hero copy on runlegion.dev (filed separately once this ships).

ssilvius:
## Summary

Closes #240.

Three subcommands under \`legion watch\` for managing \`watch.toml\` from the CLI:

- \`legion watch add <path> [--name <name>] [--agent <agent>]\` -- idempotent by canonicalized path; name defaults to path basename; \`--agent\` decouples the signal recipient from the project name (e.g., the \`platform\` agent watching the \`ledger\` repo).
- \`legion watch remove <name>\`
- \`legion watch list\`

Bare \`legion watch\` keeps its daemon behavior unchanged.

Closes the credibility gap in the runlegion.dev hero -- \`legion watch\` could register repos from the CLI in the old Bun server and got dropped during the Rust port.

## Schema changes

\`WatchRepoConfig\` gains:

- Optional \`agent: Option<String>\` -- serde default + skip_serializing_if, backward compatible.
- \`extra: toml::Table\` via \`#[serde(flatten)]\` -- captures any per-repo fields the struct does not model (\`github\`, \`worksource\`, \`review\`, future fields) so CRUD never strips them. Regression test guards this.

## Signal routing

\`WatchRepoConfig::recipient()\` centralizes \"prefer agent, else name.\" \`poll_cycle\` uses it for @mention pattern matching. Self-exclusion and \`watch_handled\` lookups remain keyed on \`repo.name\` so:

- An agent \`platform\` watching the \`ledger\` repo does not wake on signals it posted itself.
- Editing \`agent =\` later does not replay previously handled signals.

## Review pass (pr-review-toolkit + /legion-simplify)

- **Self-signal routing**: \`get_unhandled_signals_for_repo\` now takes separate \`repo_name\` and \`recipient\`.
- **Empty-agent bypass**: \`agent = \"\"\` or whitespace in hand-edited config made \`recipient()\` return empty. Now trimmed/filtered; \`add_repo_to_config\` normalizes to \`None\`.
- **Data-loss footgun**: typed serde round-trip was dropping unknown fields. Fixed via flatten. Regression test seeds \`github =\` and asserts it survives.
- **IO/TOML error context**: all \`fs\` and \`toml\` calls wrap errors with operation and path.
- **Auto-unblock silent-failure**: now logs instead of swallowing signal-lookup errors.
- Simplify: \`list_repos_in_config\` delegates to \`read_or_default\`; idempotency is canonical-path only per spec; name collision returns an explicit error.

## Test plan

- [x] \`cargo test\` -- 511 tests pass (419 unit + 9 + 83 integration).
- [x] \`cargo clippy -- -D warnings\` and \`cargo fmt -- --check\` clean.
- [x] Manual smoke: add/remove/list work; bare \`legion watch\` unchanged.
- [x] Unknown-field preservation verified on real \`watch.toml\`: 14 \`github\` entries before and after round-trip.
- [x] Backward compat: configs without \`agent\` load and serialize without drift.

## Commits

- \`05d09a1\` -- initial subcommands, config schema, daemon routing change.
- \`831ee10\` -- \`--name\` optional, derive from basename.
- \`4acbb2f\` -- simplify pass: \`recipient()\` method, canonical-path-only idempotency, \`list_repos_in_config\` delegation.
- \`56da4b0\` -- review pass: self-signal routing, empty agent normalization, IO context, silent-failure logging.
- \`d759a99\` -- preserve unknown per-repo fields on read-modify-write (data-loss fix).

## Out of scope

Hero-copy update to runlegion.dev; \`watch remove\` by agent/path; \`watch edit\` subcommand; multi-agent-per-repo.

### DIFF
--- a/plugin/commands/snooze.md
+++ b/plugin/commands/snooze.md
@@ -31,8 +31,10 @@ Every boost makes the system smarter. Do not skip this.
 Store a session reflection that captures what matters for future sessions. Focus on WHY, not WHAT:
 
 ```bash
-legion reflect --repo <your-repo> --text "<consolidated session summary>"
+legion reflect --repo <your-repo> --text "<consolidated session summary>" --domain snooze --tags session
 ```
+
+The `--domain snooze` tag is mandatory -- SessionStart pulls the most recent `--domain snooze` reflection on boot. Without the tag, your session summary is invisible to the next session.
 
 Good reflections answer: "What would I tell another agent who hits this same situation tomorrow?"
 
--- a/src/db.rs
+++ b/src/db.rs
@@ -1071,10 +1071,15 @@ impl Database {
     pub fn get_unhandled_signals_for_repo(
         &self,
         repo_name: &str,
+        recipient: &str,
         since: Option<&str>,
     ) -> Result<Vec<Reflection>> {
-        let pattern_start = format!("@{} %", repo_name);
-        let pattern_mid = format!("%@{} %", repo_name);
+        // `recipient` drives the @mention pattern matching (this is the agent name
+        // when configured, otherwise the repo name). `repo_name` is the stable key
+        // for the watch_handled table and for self-signal exclusion -- signals are
+        // stored by their source repo, not by any agent override.
+        let pattern_start = format!("@{} %", recipient);
+        let pattern_mid = format!("%@{} %", recipient);
         let pattern_all_start = "@all %";
         let pattern_all_mid = "%@all %";
         let since_clause = if since.is_some() {
--- a/src/main.rs
+++ b/src/main.rs
@@ -3639,7 +3639,7 @@ fn run() -> error::Result<()> {
                                 .map(str::to_string)
                                 .ok_or_else(|| {
                                     error::LegionError::WatchConfig(format!(
-                                        "cannot derive repo name from path {}",
+                                        "cannot derive a repo name from path {} -- pass --name <name> explicitly",
                                         path.display()
                                     ))
                                 })?
--- a/src/main.rs
+++ b/src/main.rs
@@ -1191,12 +1191,13 @@ enum QualityGateAction {
 enum WatchAction {
     /// Add a working directory to watch.toml (idempotent by canonicalized path)
     Add {
-        /// Display name for this repo (used for cooldown tracking)
-        #[arg(long)]
-        name: String,
-
         /// Absolute or relative path to the working directory
         path: std::path::PathBuf,
+
+        /// Display name for this repo (used for cooldown tracking).
+        /// Defaults to the basename of the resolved path.
+        #[arg(long)]
+        name: Option<String>,
 
         /// Signal recipient name. Signals addressed to this value (e.g., @my-agent)
         /// will wake this repo. Defaults to --name when absent.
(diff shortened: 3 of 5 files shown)


ssilvius (commented):
Review focus: rust idioms, error handling, serde flatten correctness.

## Serde flatten correctness (biggest concern)

`WatchRepoConfig.extra: toml::Table` with `#[serde(flatten)]` is the load-bearing mechanism for preserving third-party fields through CRUD. The test `add_repo_preserves_unknown_fields_on_existing_entries` only covers scalar extras (`github`, `worksource`). Nested-table extras (e.g., `[[repos]]` with a `review = { ... }` inline table, or a `[repos.matcher]` sub-table) are a known weak spot for serde flatten + toml-rs -- they can fail to round-trip or produce ambiguous output.

Recommend adding a regression test that seeds a nested-table extra and asserts it survives a round-trip through add/remove. If it fails, we need to either document "scalar extras only" or switch the type (e.g., `HashMap<String, toml::Value>` with explicit nested-handling). Cheap to check now, expensive to find via someone's `legion pr create` integration blowing up later.

`skip_serializing_if = "toml::Table::is_empty"` on a flattened field is a no-op in practice (empty tables emit nothing anyway), but harmless.

## Error handling

**`poll_cycle` inconsistency** (src/watch.rs): the auto-unblock pass logs-and-continues on `find_pending_signals` failure, but the main delivery pass propagates via `?`. If the query is genuinely broken, every cycle will emit a warning then immediately abort. Pick one strategy -- either both pass errors up, or both log-and-continue per repo so one bad repo does not sink the whole poll.

**Lossy UTF-8 for stored workdir** (src/watch.rs add_repo_to_config): `canonical.to_string_lossy().into_owned()` silently drops non-UTF8 bytes. Since this string is the idempotency key, a lossy conversion means a second add for the same (non-UTF8) path would not match and would insert a duplicate. Inconsistent with main.rs name derivation, which correctly uses `to_str().ok_or_else(...)`. Recommend strict conversion in `add_repo_to_config` too.

**Agent display layer** (src/main.rs Watch::Add handler): when `--agent ""` is passed, the println prints `(agent: )` even though `add_repo_to_config` normalizes it to None. Read the stored value back (or normalize before formatting the message) so output matches on-disk state.

## Rust idioms

**Double canonicalize**: main.rs canonicalizes for name derivation, then `add_repo_to_config` canonicalizes again. Non-functional, but if the two error messages ever diverge (they already phrase differently: "cannot be resolved" vs "is not a directory") users get confusing output depending on which failure fires first. Consider canonicalizing once and threading the PathBuf through.

**`recipient()` asymmetry**: trims the agent override but returns `&self.name` unmodified. Not a bug given name sources are controlled, but technically asymmetric. Trivial.

## Tests

Solid coverage of the happy paths plus the two regression guards I care about (self-signal exclusion via repo_name; empty-agent normalization). Missing: nested-table flatten round-trip (see above). That is the one I would actually block on.

Not blocking approval from the rust-idioms angle -- the serde flatten test gap is worth resolving but the rest are polish. Approve contingent on either adding the flatten round-trip test or acknowledging the scalar-only limitation.

Issue #277, opened 2026-04-21
feat: TCP listener so peers can push sync connections into this node


ssilvius:
## Goal

Add an incoming TCP listener to the sync actor so that remote peers can open connections into this node and push deltas via ``smugglr_core::broadcast::handle_sync_connection``. Without this, the actor is one-way only: a node can initiate sync outbound, but it can never be reached by a peer that initiates. Bidirectional replication requires both halves.

## Context

Smugglr's broadcast protocol is two-sided. Each node advertises a ``sync_port`` in its UDP announcement payload; peers connect TCP to that port when they want to sync. ``smugglr_core::broadcast::handle_sync_connection`` (at ``crates/smugglr-core/src/broadcast.rs:881``) is the server-side handler: it reads a ``SyncRequest`` off the stream, computes the response (deltas + ``want_rows``), writes the ``SyncResponse``, then reads follow-up ``DeltaPacket``s and applies them via ``LocalDb::upsert_rows``.

Legion's current sync actor (``src/sync_actor.rs``) binds no listener. Peers discover this node's existence via UDP announcements but have nowhere to connect. Combined with the outgoing-side gap (companion issue), this is why ``legion cluster status`` shows ``last_sync: never`` even when two nodes see each other.

### Interface

Spawn a tokio task alongside the outgoing loop inside ``SyncHandle::spawn`` (or inside ``run_sync_loop``):

``````rust
async fn run_sync_listener(
    sync_port: u16,
    smugglr_config: Arc<smugglr_core::Config>,
    broadcast_config: Arc<smugglr_core::broadcast::BroadcastConfig>,
    replay_guard: Arc<tokio::sync::Mutex<smugglr_core::broadcast::ReplayGuard>>,
    mut stop_rx: tokio::sync::watch::Receiver<bool>,
) -> Result<()> {
    let listener = tokio::net::TcpListener::bind(("0.0.0.0", sync_port)).await?;
    loop {
        tokio::select! {
            _ = stop_rx.changed() => break,
            accept = listener.accept() => {
                let (stream, peer) = accept?;
                let cfg = smugglr_config.clone();
                let bc = broadcast_config.clone();
                let guard = replay_guard.clone();
                tokio::spawn(async move {
                    if let Err(e) = smugglr_core::broadcast::handle_sync_connection(
                        stream, &cfg, &bc, &guard,
                    ).await {
                        eprintln!("[legion sync] incoming connection from {peer} failed: {e}");
                    }
                });
            }
        }
[...]

Issue #276, opened 2026-04-21
feat: replace wall-clock delta query with smugglr run_broadcast_once


ssilvius:
## Goal

Delete the hand-rolled wall-clock delta query from the outgoing sync path and delegate the entire outbound sync cycle to ``smugglr_core::broadcast::run_broadcast_once``. Time-based ``get_*_deltas_since(&last_sync)`` queries make sync decisions on wall-clock comparisons, which is the anti-pattern smugglr was built to avoid. Content-hash reconciliation is smugglr's contract; legion must use it, not reimplement around it.

## Context

``src/sync_actor.rs`` currently keys the outgoing sync on ``last_sync = chrono::Utc::now().to_rfc3339()`` and fetches deltas via ``db.get_reflection_deltas_since(&last_sync)`` + ``get_card_deltas_since`` + ``get_schedule_deltas_since``. The collected deltas are then logged with a ``TODO: Actually broadcast`` comment (line 155). No data is actually transmitted.

This path is wrong on two axes:

1. It does not reach the wire at all.
2. Even if it did, keying the query on wall-clock time violates smugglr's sync contract. Clocks skew, clocks lie, NTP drifts. UUIDv7 is used for post-hoc ordering only; it is never read to make sync decisions. Reflection ``019d7c22`` (PR #222 lesson) is explicit: no ``Utc::now()`` in any cursor or query that drives replication.

Smugglr's public surface already provides the correct primitive. ``smugglr_core::broadcast::run_broadcast_once`` (at ``crates/smugglr-core/src/broadcast.rs:1060``) performs the entire outbound cycle: peer discovery, TCP connect, ``SyncRequest`` with per-row content hashes, ``SyncResponse`` with deltas, apply via ``LocalDb::upsert_rows``. Content-addressed, idempotent, bidirectional.

Reflection ``019d7b24`` confirmed for legion v1: inner-path access (``smugglr_core::broadcast::*``) is expected, ``ReplayGuard`` in-memory is sufficient (legion daemon is long-lived), no direct ``LocalDb::upsert_rows`` needed since smugglr owns that call inside ``run_broadcast_once``.

### Interface

Replace the outgoing cycle in ``src/sync_actor.rs::run_sync_loop``:

``````rust
// Remove:
let reflection_deltas = db.get_reflection_deltas_since(&last_sync)...;
let card_deltas = db.get_card_deltas_since(&last_sync)...;
let schedule_deltas = db.get_schedule_deltas_since(&last_sync)...;
[...]

PR #288, opened 2026-04-21
feat(#287): legion statusline subcommand -- rate-limit + usage ingest, chip render

### LINKED ISSUE #287: feat: legion statusline subcommand -- Claude Code statusLine ingest + display
## Goal

Add a ``legion statusline`` subcommand that consumes the Claude Code statusLine JSON on stdin, persists rate-limit and usage samples to legion's local store, and renders a single-line status chip to stdout. This is the ingest-side foundation for the pillar-2 scheduler/gate work: every assistant turn writes authoritative rate-limit state (``rate_limits.five_hour`` / ``rate_limits.seven_day``) into a cluster-syncable table, giving every node in the cluster a live view of account-wide headroom.

The shape is locked from two sources:
- Claude Code statusline schema at ``code.claude.com/docs/en/statusline`` (official JSON contract).
- Mark Turrell's MIT statusline at ``github.com/mtberlin2023/claude-code-skills/tree/main/statusline`` (reference for thresholds, render format, synthetic-content filters, cache-window pattern). NOT a dependency. Legion ships native Rust.

## Context

Reflection ``019daed3-13ec`` captures the distilled shape. Short version:

- stdin JSON: ``transcript_path``, ``session_id``, ``model.id``, ``rate_limits.five_hour.{used_percentage,resets_at}``, ``rate_limits.seven_day.{used_percentage,resets_at}``, ``cost.total_cost_usd``, ``context_window.used_percentage``.
- Per-turn replay = most recent assistant message's ``usage.cache_read_input_tokens + cache_creation_input_tokens + input_tokens``.
- Synthetic-content prefixes to skip when counting real user turns: ``<system-reminder>``, ``<command-message>``, ``<command-name>``, ``<command-args>``, ``<task-notification>``, ``<local-command-caveat>``, ``<user-prompt-submit-hook>``, ``Unknown command:``, ``Caveat:``.
- Cached ``tool_result`` >2KB with ``is_error: true`` = surfaceable error signal.
- Display thresholds (starting calibration, refine from data):
  - 5h cap: warn 75%, alert 90%.
  - Weekly cap: warn 60%, alert 85%.
  - Next-turn replay: green <200K, yellow 200-499K, red 500-999K, hard-cap red >=1M.
- Cache rendered line at ``/tmp/legion-statusline-<session_id>`` for 10s.

Legion's ``src/usage.rs`` already parses transcript JSONL for ``message.usage``. Reuse that code path; do not duplicate the aggregation.

### Interface

New subcommand:

``````rust
// src/main.rs enum Commands
Statusline {
    /// Output as JSON instead of the rendered chip (for testing / scripting)
    #[arg(long)]
    json: bool,
}
``````

Reads JSON on stdin. Writes one row per invocation to ``rate_limit_samples`` and (when a transcript path is present and usage data is readable) one row to ``usage_samples``. Prints the rendered chip on stdout. Cache-hits short-circuit both DB writes and re-render.

### Schema -- new tables

``````sql
CREATE TABLE rate_limit_samples (
    id TEXT PRIMARY KEY,               -- UUIDv7
    hostname TEXT NOT NULL,
    session_id TEXT NOT NULL,
    sampled_at TEXT NOT NULL,          -- ISO8601
    five_hour_pct REAL,                -- 0-100, nullable if absent
    five_hour_resets_at INTEGER,       -- unix epoch, nullable
    seven_day_pct REAL,                -- 0-100, nullable
    seven_day_resets_at INTEGER,       -- unix epoch, nullable
    model TEXT,                        -- model.id, nullable
    deleted_at TEXT,                   -- smugglr sync
    updated_at TEXT                    -- smugglr sync
);
CREATE INDEX idx_rate_limit_samples_sampled ON rate_limit_samples(sampled_at);
CREATE INDEX idx_rate_limit_samples_host ON rate_limit_samples(hostname);

CREATE TABLE usage_samples (
    id TEXT PRIMARY KEY,               -- UUIDv7
    hostname TEXT NOT NULL,
    session_id TEXT NOT NULL,
    turn_index INTEGER,                -- nth real user turn in session, nullable on error
    model TEXT,
    input_tokens INTEGER NOT NULL DEFAULT 0,
    output_tokens INTEGER NOT NULL DEFAULT 0,
    cache_write_tokens INTEGER NOT NULL DEFAULT 0,
    cache_read_tokens INTEGER NOT NULL DEFAULT 0,
    effective_tokens INTEGER NOT NULL,   -- computed at write per src/usage.rs formula
    error_bytes INTEGER NOT NULL DEFAULT 0, -- max cached tool_result error size in bytes
    sampled_at TEXT NOT NULL,
    deleted_at TEXT,
    updated_at TEXT
);
CREATE INDEX idx_usage_samples_sampled ON usage_samples(sampled_at);
CREATE INDEX idx_usage_samples_host ON usage_samples(hostname);
CREATE INDEX idx_usage_samples_session ON usage_samples(session_id);
``````

Both tables are smugglr-syncable. Add delta serialization + sync declarations matching existing reflections/tasks/schedules pattern.

### Behavior

- Reads stdin JSON. Parses via ``serde_json::Value``. Missing or malformed fields do NOT abort; nullable columns stay NULL. Silent on partial data.
- ``hostname`` resolved via existing legion helper (match what health samples use).
- Per-session cache: ``/tmp/legion-statusline-<session_id>`` mtime check. If younger than 10s, print cached content and exit 0 without DB writes.
- Transcript parsing reuses ``src/usage.rs``. Add a thin wrapper that returns the latest assistant turn's token counts plus the max error size across recent ``tool_result`` items.
- Real-turn counting: iterate user messages, filter synthetic prefixes, return ``turn_index`` for the most recent real turn.
- Render chip format matches mtberlin2023 pattern:
  ``<emoji> [<turns>] <tokens> next turn[ · <action> after task][ · <action> now][ · <action> before next msg][ · WARN <N>KB err][ · WARN/ALERT <pct>%·5h->...][ · WARN/ALERT <pct>%·wk->...]``
  Weekly reset formatted as ``"Day HH:MM"`` local time; 5h reset as ``HH:MM`` same-day or ``"Day HH:MM"`` other-day. No emoji in source other than the documented chip characters.
- Env override: ``LEGION_STATUSLINE_ACTION`` (default ``/log``) controls the action hint at yellow+ replay thresholds.
- On any unexpected failure (DB write error, parse error past recovery), print nothing, exit 0. A statusline that breaks the Claude Code UI is worse than no chip.

### Error Handling

- Stdin not JSON -> exit 0 silently.
- Missing ``session_id`` -> exit 0 silently; cannot cache or key rows.
- DB open/write failure -> print chip anyway (best-effort ingest), log to ``/tmp/legion-hook-errors.log``.
- Transcript file not found / unreadable -> skip usage sample, render chip with rate-limit data only.
- No emoji in code except inside the chip-render literal strings.

## Out of Scope

- ``setup-binary.sh`` integration that writes ``statusLine.command`` into user settings (separate issue).
- ``legion budget show`` / ``legion budget check`` query surface for the gate (separate issue).
- Weekly runway chip (depends on cluster-wide usage_samples having enough data; follow-up issue).
- Per-model weight calibration (opus-vs-sonnet multiplier derived from paired observations; separate issue).
- Replacing ``src/usage.rs`` aggregation formula; this issue reuses it as-is.

## File Locations

- Schema: ``src/db.rs`` (new migration).
- Subcommand: ``src/statusline.rs`` (new module).
- CLI wiring: ``src/main.rs`` (new ``Statusline`` variant on ``Commands`` + dispatch).
- Delta types: ``src/sync.rs`` (new ``RateLimitDelta``, ``UsageDelta``; registered alongside existing ``ReflectionDelta`` / ``CardDelta`` / ``ScheduleDelta``).
- Tests: bottom of ``src/statusline.rs`` in ``#[cfg(test)]`` using hand-crafted stdin JSON fixtures + tempdir DB.

## Done When

- [ ] ``rate_limit_samples`` and ``usage_samples`` migrations land.
- [ ] Smugglr delta types for both tables; both declared syncable in the sync actor config surface.
- [ ] ``legion statusline`` subcommand runs end-to-end: accepts Claude Code JSON on stdin, writes a ``rate_limit_samples`` row, writes a ``usage_samples`` row when transcript is readable, prints the rendered chip.
- [ ] 10-second per-session cache short-circuits re-render and DB writes.
- [ ] Render matches the documented format. Five unit tests cover: healthy (no caps shown), yellow replay, red replay, 5h-warn, weekly-alert.
- [ ] Synthetic-content filter prevents fake user turns from incrementing ``turn_index``. Unit test with all nine synthetic prefixes.
- [ ] ``--json`` flag prints the parsed + computed values as JSON instead of the chip (for testing and downstream consumers).
- [ ] ``cargo test``, ``cargo clippy -- -D warnings``, ``cargo fmt -- --check`` all pass.
- [ ] Simplify pass completed; review pass completed and issues fixed.
- [ ] PR references this issue.

**This issue is complete when:** a Claude Code session configured with ``statusLine.command = "legion statusline"`` produces the expected chip on every render, and ``rate_limit_samples`` + ``usage_samples`` rows are written with each sample, visible to other cluster nodes once smugglr sync is wired.

## Context

- Companion work (separate issues): setup-binary.sh integration for ``statusLine.command`` write; ``legion budget`` query surface; runway chip; gate wiring into ``legion work`` and watch spawn.
- Reference: reflection ``019daed3-13ec`` (distilled shapes, thresholds, install pattern).
- Related: ``src/usage.rs`` (transcript parsing reuse); multi-node sync issues #276 (outgoing) and #277 (listener) -- the tables declared here become syncable once those land.
- Smugglr content-hash doctrine: no wall-clock-keyed queries in the sync path. Delta types follow the existing pattern.

ssilvius:
## Summary

Closes #287. Adds the `legion statusline` subcommand: Claude Code pipes its
statusLine JSON to stdin, legion persists the rate-limit and usage samples
to its local store, renders a single-line chip on stdout.

This is the pillar-2 foundation: every assistant turn writes authoritative
5-hour and 7-day rate-limit state into a cluster-syncable table, giving
every node in the cluster a live view of account-wide headroom. The
budget gate wires onto this surface in a follow-up.

## What's in the diff

- **Schema (migration 16)**: `rate_limit_samples` and `usage_samples`
  tables. Both carry `deleted_at` + `updated_at` for smugglr content-hash
  sync. Partial indexes on live rows.
- **`legion statusline`**: parses the Claude Code JSON via `serde_json::Value`
  (robust to partial payloads), reuses `src/usage.rs` token accounting via
  a new `parse_transcript_tail`, writes samples, renders the chip, caches
  output at `/tmp/legion-statusline-<session>` for 10s.
- **Chip format**: matches the proven mtberlin2023 MIT reference. Emoji by
  next-turn-replay band (green/yellow/red/hard-cap), action hint at yellow+,
  cached `tool_result` error warning >2KB, 5h/weekly cap segments with local
  reset times. Chip glyphs are unicode escapes so the source stays ASCII.
- **Thresholds** (starting calibration; refine from data):
  - replay: green <200K, yellow 200-499K, red 500-999K, hard-cap >=1M
  - 5h: warn 75%, alert 90%
  - weekly: warn 60%, alert 85%
- **Smugglr deltas**: `RateLimitDelta`, `UsageDelta` with `from_sample`
  constructors. Wired into sync actor once #276 lands transport.
- **Transcript tail**: `parse_transcript_tail` in `usage.rs` filters nine
  synthetic content prefixes so real user turns are counted correctly.

## Test plan

- [x] 21 new unit tests in `src/statusline.rs` cover band thresholds,
      cap segments, error suffix, cache freshness, input parsing (full,
      partial, garbage), path sanitisation, render end-to-end for healthy
      and critical states.
- [x] 3 new `sync.rs` tests cover delta round-trip for rate-limit and
      usage samples including tombstone paths.
- [x] `cargo test` -- 87 passing, 0 failures, 0 regressions.
- [x] `cargo clippy -- -D warnings` clean.
- [x] `cargo fmt --check` clean.
- [x] Smoke-tested end-to-end with a real Claude Code transcript; chip
      renders as expected: `🟡 [34] 409K next turn · /log after task
      · 🚨 91%·5h→Tue 11:00 · 🚨 88%·wk→Sun 02:06`.
- [x] `/legion-simplify` gate: clean, recorded via `legion quality-gate record`.
- [x] Pre-push review agent: all findings addressed (pct.round(), JSON
      error swallow, cache-path comment).

## Out of scope

These are follow-up issues, not blockers:

- `setup-binary.sh` integration that writes `statusLine.command` into
  `~/.claude/settings.json` idempotently on SessionStart.
- `legion budget show` / `legion budget check` query surface for the gate.
- Weekly runway chip (depends on cluster-wide `usage_samples` accumulating
  enough data; needs #276 + #277 transport first).
- Per-model weight calibration (sonnet-1x / opus-multiplier derived from
  paired observations).

## Reference

Mark Turrell's MIT statusline at github.com/mtberlin2023/claude-code-skills
informed the threshold calibration and chip render conventions. Not a
fork or dependency; legion's implementation is native Rust.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -581,6 +581,56 @@ impl Database {
                  ON schedules(repo) WHERE deleted_at IS NULL;",
         )?;
 
+        // Migration 16: Rate limit and usage samples for pillar-2 scheduler (#287).
+        // Populated by `legion statusline` on every Claude Code render. Both
+        // tables carry deleted_at + updated_at for smugglr content-hash sync.
+        conn.execute_batch(
+            "CREATE TABLE IF NOT EXISTS rate_limit_samples (
+                id TEXT PRIMARY KEY,
+                hostname TEXT NOT NULL,
+                session_id TEXT NOT NULL,
+                sampled_at TEXT NOT NULL,
+                five_hour_pct REAL,
+                five_hour_resets_at INTEGER,
+                seven_day_pct REAL,
+                seven_day_resets_at INTEGER,
+                model TEXT,
+                deleted_at TEXT,
+                updated_at TEXT
+            );
+            CREATE INDEX IF NOT EXISTS idx_rate_limit_samples_sampled \
+                ON rate_limit_samples(sampled_at);
+            CREATE INDEX IF NOT EXISTS idx_rate_limit_samples_host \
+                ON rate_limit_samples(hostname);
+            CREATE INDEX IF NOT EXISTS idx_rate_limit_samples_live \
+                ON rate_limit_samples(hostname, sampled_at) WHERE deleted_at IS NULL;
+
+            CREATE TABLE IF NOT EXISTS usage_samples (
+                id TEXT PRIMARY KEY,
+                hostname TEXT NOT NULL,
+                session_id TEXT NOT NULL,
+                turn_index INTEGER,
+                model TEXT,
+                input_tokens INTEGER NOT NULL DEFAULT 0,
+                output_tokens INTEGER NOT NULL DEFAULT 0,
+                cache_write_tokens INTEGER NOT NULL DEFAULT 0,
+                cache_read_tokens INTEGER NOT NULL DEFAULT 0,
+                effective_tokens INTEGER NOT NULL,
+                error_bytes INTEGER NOT NULL DEFAULT 0,
+                sampled_at TEXT NOT NULL,
+                deleted_at TEXT,
+                updated_at TEXT
+            );
+            CREATE INDEX IF NOT EXISTS idx_usage_samples_sampled \
+                ON usage_samples(sampled_at);
+            CREATE INDEX IF NOT EXISTS idx_usage_samples_host \
+                ON usage_samples(hostname);
+            CREATE INDEX IF NOT EXISTS idx_usage_samples_session \
+                ON usage_samples(session_id);
+            CREATE INDEX IF NOT EXISTS idx_usage_samples_live \
+                ON usage_samples(hostname, sampled_at) WHERE deleted_at IS NULL;",
+        )?;
+
         Ok(())
     }
 
@@ -2350,6 +2400,86 @@ impl Database {
             [older_than],
         )?;
         Ok(rows as u64)
+    }
+
+    // -- Statusline Samples ------------------------------------------------------
+
+    /// Insert a rate-limit sample captured from a Claude Code statusline render.
+    pub fn insert_rate_limit_sample(
+        &self,
+        sample: &crate::statusline::RateLimitSample,
+    ) -> Result<String> {
+        self.conn.execute(
+            "INSERT INTO rate_limit_samples (id, hostname, session_id, sampled_at, \
+             five_hour_pct, five_hour_resets_at, seven_day_pct, seven_day_resets_at, \
+             model, updated_at) \
+             VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?4)",
+            rusqlite::params![
+                sample.id,
+                sample.hostname,
+                sample.session_id,
+                sample.sampled_at,
+                sample.five_hour_pct,
+                sample.five_hour_resets_at,
+                sample.seven_day_pct,
+                sample.seven_day_resets_at,
+                sample.model,
+            ],
+        )?;
+        Ok(sample.id.clone())
+    }
+
+    /// Insert a usage sample captured from a Claude Code statusline render.
+    pub fn insert_usage_sample(&self, sample: &crate::statusline::UsageSample) -> Result<String> {
+        self.conn.execute(
+            "INSERT INTO usage_samples (id, hostname, session_id, turn_index, model, \
+             input_tokens, output_tokens, cache_write_tokens, cache_read_tokens, \
+             effective_tokens, error_bytes, sampled_at, updated_at) \
+             VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11, ?12, ?12)",
+            rusqlite::params![
+                sample.id,
+                sample.hostname,
+                sample.session_id,
+                sample.turn_index,
+                sample.model,
+                sample.input_tokens,
+                sample.output_tokens,
+                sample.cache_write_tokens,
+                sample.cache_read_tokens,
+                sample.effective_tokens,
+                sample.error_bytes,
+                sample.sampled_at,
+            ],
+        )?;
+        Ok(sample.id.clone())
+    }
+
+    /// Most recent rate-limit sample across all hosts in the cluster.
+    /// Returns `None` when no sample has been written yet. Used by the
+    /// budget gate to read cluster-wide headroom.
+    #[allow(dead_code)] // Consumed by the upcoming `legion budget` subcommand.
+    pub fn latest_rate_limit_sample(&self) -> Result<Option<crate::statusline::RateLimitSample>> {
+        let mut stmt = self.conn.prepare(
+            "SELECT id, hostname, session_id, sampled_at, five_hour_pct, \
+             five_hour_resets_at, seven_day_pct, seven_day_resets_at, model \
+             FROM rate_limit_samples WHERE deleted_at IS NULL \
+             ORDER BY sampled_at DESC LIMIT 1",
+        )?;
+        let mut rows = stmt.query([])?;
+        match rows.next().map_err(LegionError::Database)? {
+            Some(row) => Ok(Some(crate::statusline::RateLimitSample {
+                id: row.get(0)?,
+                hostname: row.get(1)?,
+                session_id: row.get(2)?,
+                sampled_at: row.get(3)?,
+                five_hour_pct: row.get(4)?,
+                five_hour_resets_at: row.get(5)?,
+                seven_day_pct: row.get(6)?,
+                seven_day_resets_at: row.get(7)?,
+                model: row.get(8)?,
+            })),
+            None => Ok(None),
+        }
     }
 
     /// Rename a repo across all tables. Returns total rows updated.
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,6 +18,7 @@ mod serve;
 mod signal;
 mod stats;
 mod status;
+mod statusline;
 mod surface;
 mod sync;
 mod sync_actor;
@@ -536,6 +537,18 @@ enum Commands {
         all_hosts: bool,
 
         /// Output as JSON
+        #[arg(long)]
+        json: bool,
+    },
+
+    /// Claude Code statusLine subcommand: ingest rate-limit + usage, render chip
+    ///
+    /// Reads the Claude Code statusLine JSON on stdin, persists a
+    /// `rate_limit_samples` row (and a `usage_samples` row when the
+    /// transcript is readable), then prints a single-line chip on stdout.
+    /// Wire via `statusLine.command` in settings.json.
+    Statusline {
+        /// Print the parsed sample state as JSON instead of the chip
         #[arg(long)]
         json: bool,
     },
@@ -3835,6 +3848,12 @@ fn run() -> error::Result<()> {
                 println!("{}", row.id);
             }
         },
+        Commands::Statusline { json } => {
+            // Never surface an error to the Claude Code UI: statusline::run
+            // already swallows internal failures and logs them.
+            statusline::run(json)?;
+        }
+
         Commands::Usage {
             session,
             since,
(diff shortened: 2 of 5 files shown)


ssilvius (commented):
@legion -- review:rust-idioms-correctness for #288

Overall: the shape is right -- one module owning the read/persist/render flow, statusline never panics or surfaces an error to Claude Code UI, samples cleanly typed and persisted, sync deltas paired in the same PR. Test coverage hits replay bands, cap segments, parse_input, cache sanitisation, build_usage_sample contract, and DB round-trips. I would land this with the items below addressed (or at least filed as follow-ups).

## Real correctness issues

**1. `write_cache` is non-atomic; cache hit can read a partially-written chip.**
`statusline.rs:447-453` does `File::create` then `write_all`. `File::create` truncates synchronously; `write_all` may be observed mid-flight by a concurrent statusline render that calls `read_cache_if_fresh` within the 10s TTL. The reader will see a fresh mtime + truncated bytes and emit a partial chip.

The "cached content already carries its trailing newline ... emit it verbatim with print! rather than println!" invariant in the run() comment depends on the cache file containing the full bytes, which create+write_all does not guarantee under concurrent renders for the same session_id.

Fix: write to `<path>.tmp` then atomic rename. `tempfile::NamedTempFile::persist` is already in your dev-deps for tests.

**2. `/tmp` hardcoding leaks across users + breaks Windows.**
`CACHE_PREFIX = "/tmp/legion-statusline-"` and `ERROR_LOG = "/tmp/legion-hook-errors.log"` are POSIX-only and world-readable on shared hosts. The DB path uses `directories::ProjectDirs` correctly -- runtime artefacts should too. `ProjectDirs::cache_dir()` and `data_dir()` give you the right per-user, per-platform locations.

Cache poisoning angle: on a multi-user box, another user can pre-create `/tmp/legion-statusline-<known-session-id>` with attacker-chosen bytes; sanitiser doesn't help because the sanitiser runs on your input, not on the existing file. Same TOCTOU on the error log via append.

**3. u64 -> i64 truncating casts on token counts.**
`statusline.rs:262-266` and similar: `raw.input as i64`, etc. SQLite's INTEGER is i64 so the cast is structurally required, but `as` truncates silently above 2^63. Practically safe (token counts will never approach this), but `i64::try_from(raw.input).unwrap_or(i64::MAX)` removes the silent-overflow footgun and is the idiomatic pattern when you know an upper bound exists. Same with `tail.real_turns as i64` and the i64 effective_tokens conversion.

## Patterns / style

**4. `pub fn run(json: bool) -> Result<()>` that can never return Err.**
The doc comment explicitly says "Always returns Ok(()) in production". The call site in `main.rs:3851` has a no-op `?`. Either drop the Result -- `pub fn run(json: bool)` -- or actually surface a typed error variant for the JSON-mode path. Today's signature lies to the caller about its failure modes.

**5. `INSERT ... VALUES (?1, ..., ?9, ?4)` reuses bind index for `updated_at`.**
`db.rs` insert_rate_limit_sample and insert_usage_sample reuse the sampled_at bind for updated_at. Cute and correct today, but a future maintainer adding an UPDATE path will not get the same coupling. One-line comment above the SQL stating "updated_at intentionally bound to sampled_at on insert; UPDATE callers must rebind" would prevent that footgun.

**6. Doc-drift in `parse_transcript_tail`.**
`usage.rs:6-8` says "one stderr warning per session, not per line" -- that aggregation is in `SessionUsage::from_file`, not `parse_transcript_tail`, which silently `continue`s on every malformed line. Worth either adding the same skip_count + final eprintln! to parse_transcript_tail or trimming the doc to match.

**7. `is_real_user_turn` falls through to `real = true` on unknown content types.**
`usage.rs:697-699`. Bias is toward over-counting which is the safer direction here, but worth a `// any unknown content-item type counts as a real turn (bias to over-count rather than miss real input)` comment so it's an explicit choice rather than a fallthrough.

## Smaller things

**8. `sysinfo::System::host_name()`** drags a heavy crate for a hostname. `gethostname` crate is ~15 lines and has no transitive deps. Skip if sysinfo is already in your dependency tree for other reasons.

**9. `read_to_string(&mut raw)` on stdin** has no length cap. Claude Code closes stdin so practically fine, but a 1MB upper bound via `take(1 << 20).read_to_string(...)` would prevent a future hostile or buggy hook from exhausting RAM.

**10. Test gap**: no roundtrip assertion that `write_cache` + `read_cache_if_fresh` yields byte-identical output to the live render. The "cached content already carries its trailing newline" invariant is load-bearing for the print! vs println! choice; one test would lock it in.

## Things that look good

- `next_turn_replay` excluding output tokens with the explicit "Output is deliberately excluded" doc -- correct semantic and well-explained.
- Rate-limit / usage delta types landed alongside their schemas in the same PR; sync wire-up in #276 will not be blocked.
- `build_usage_sample` returning None when transcript absent OR all tokens zero -- right invariant for cluster aggregates.
- Cap-segment refactor (CapConfig) collapses the 5h/wk duplication cleanly.
- Migration 16 uses `CREATE TABLE IF NOT EXISTS` and leaves room for the `(hostname, sampled_at)` partial index that the budget gate will want.

Recommend: address #1 and #2 before merge (real correctness + multi-user safety). #3-#7 are land-then-iterate. #8-#10 are nice-to-haves.

-- smugglr

Issue #295, opened 2026-04-22
feat: legion pr read-side surface (view, comments, reviews, checks --log-failed)


ssilvius:
## Goal

Close the read-side gap in `legion pr`. Agents can create PRs, post reviews, and merge -- but cannot read PR bodies, comments, review bodies, or failing-check logs. The build->simplify->review->fix workflow is broken at "fix" because the agent cannot read the reviews it needs to fix, and "why is CI red" requires the blocked `gh` CLI.

### Interface

CLI:

```
legion pr view     --repo <name> --number <n> [--json]
legion pr comments --repo <name> --number <n> [--json]
legion pr reviews  --repo <name> --number <n> [--json]
legion pr checks   --repo <name> --number <n> --log-failed [--json]
```

Rust API in `src/worksource.rs`:

```rust
pub struct ExternalPRDetails {
    pub number: u64,
    pub title: String,
    pub state: String,                 // OPEN | CLOSED | MERGED
    pub author: String,
    pub created_at: String,
    pub updated_at: String,
    pub body: String,
    pub head_ref_name: String,
    pub head_sha: String,
    pub base_ref_name: String,
    pub is_draft: bool,
    pub review_decision: Option<String>,
    pub mergeable: Option<String>,     // MERGEABLE | CONFLICTING | UNKNOWN
}

pub struct ExternalPRComment {
    pub id: String,
    pub author: String,
    pub created_at: String,
    pub updated_at: String,
    pub body: String,
    pub kind: String,                  // "issue" or "review"
    pub path: Option<String>,
    pub line: Option<u32>,
    pub in_reply_to_id: Option<String>,
}

pub struct ExternalPRReview {
    pub id: String,
    pub author: String,
    pub state: String,                 // APPROVED | CHANGES_REQUESTED | COMMENTED | PENDING
    pub submitted_at: Option<String>,
    pub body: String,
    pub comments: Vec<ExternalPRComment>,
}

pub fn view_pr(plugin: &str, repo: &str, n: u64) -> Result<ExternalPRDetails>;
pub fn list_pr_comments(plugin: &str, repo: &str, n: u64) -> Result<Vec<ExternalPRComment>>;
pub fn list_pr_reviews(plugin: &str, repo: &str, n: u64) -> Result<Vec<ExternalPRReview>>;
pub fn fetch_check_logs(plugin: &str, repo: &str, job_ids: &[u64]) -> Result<Vec<(u64, String)>>;
```

[...]

PR #298, opened 2026-04-22
feat(#292): call_plugin timeout + is_failing state partition

### LINKED ISSUE #292: follow-up: call_plugin timeout + is_failing test-helper symmetry
Two non-blocking items deferred from smugglr's PR #291 review (signal 019db1f8):

## 1. call_plugin has no timeout

\`worksource::call_plugin\` shells out to plugin binaries (e.g. \`gh\`) without a
timeout. A hung network call to GitHub blocks the legion command indefinitely.
Practical impact: \`legion pr checks\` in a watch-loop or CI runner can wedge
the whole loop until the operator notices.

Likely shape: pass a configurable timeout (default ~30s) into \`call_plugin\`,
SIGKILL on expiry, return \`LegionError::WorkSource(\"plugin timeout after Ns\")\`.

## 2. is_failing test-helper passing-state list is asymmetric

\`is_failing_passing_and_in_flight_states\` enumerates the passing/in-flight
states. The current contract (post PR #291 fix) is fail-closed-on-unknown:
the function lists known passing states and treats everything else as failing.

If a future PR adds a new failure variant by including it in the negation
list, the passing-state test won't catch the omission. Smugglr's suggestion:
derive the passing list as 'all observed states minus failing' so the
contract is enforced symmetrically. Concretely, store both sets as constants
in \`worksource\` and have the test verify they partition the observed
state-space.

Both are land-then-iterate quality of life fixes. PR #291 shipped the
load-bearing pieces (fail-closed gate, plugin-not-found is Err).

ssilvius:
Closes #292.

## Summary

Two smugglr follow-ups from #291, both quality-of-life on the worksource layer.

**1. Plugin timeout.** Every `call_plugin` invocation is now bounded by a wall-clock budget (default 30s, `LEGION_WS_TIMEOUT_SECS` to override, 0 to disable). On expiry the child is SIGKILLed and `LegionError::WorkSource("plugin timeout after Ns")` is returned. Matters now because `legion pr checks / view / comments / reviews` are daily-loop commands and `legion watch` invokes them indirectly -- a hung `gh` call previously wedged the loop.

Subtlety: SIGKILL is sent to the entire process group (killpg), not just the bash interpreter. Without that, a bash plugin that spawns `gh` leaves `gh` orphaned and still holding pipe fds open; reader threads then block on EOF until gh dies, defeating the timeout. Verified by a test asserting elapsed < 5s on a 60s sleep.

Reader threads spawn BEFORE wait_timeout so large-stdout plugins do not deadlock against wait() blocking on a producer that can't drain its pipe. Covered by a 256K stdout test.

**2. `is_failing` state partition.** Extracted `PR_CHECK_PASSING_STATES` and `PR_CHECK_FAILING_STATES` as module-level constants. New `pr_check_state_lists_partition` test asserts the two sets are disjoint so a future PR can't list a state in both (or neither) and silently break the merge gate. Unknown states still fail-closed as before.

## Dependencies

- `wait-timeout = "0.2"` -- Mozilla-maintained, ~500 lines, no transitive footprint.
- `libc = "0.2"` -- unix-only target dep, used for `killpg`. Already transitive; now direct.

## Tests

5 new timeout tests (gated on `#[cfg(unix)]` since process_group/killpg are POSIX):
- stdout on success
- stderr surfaced on non-zero exit
- timeout fires + subtree killed + elapsed < 5s on 60s sleep
- no-timeout mode still works
- large stdout (256K) does not deadlock

1 new partition test guards the two state constants.

**98 total tests passing**, 0 regressions. `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

## Release posture

Scope-bundled with #295/#297 for v0.9.3 per the "daily loop" theme. Do not merge to main without explicit user approval.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1454,6 +1454,7 @@ dependencies = [
  "dirs 5.0.1",
  "hex",
  "hostname",
+ "libc",
  "mime_guess",
  "model2vec-rs",
  "rand 0.9.2",
@@ -1472,6 +1473,7 @@ dependencies = [
  "toml",
  "tower-http",
  "uuid",
+ "wait-timeout",
 ]
 
 [[package]]
@@ -3522,6 +3524,15 @@ name = "version_check"
 version = "0.9.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0b928f33d975fc6ad9f86c8f283853ad26bdd5b10b7f1542aa2fa15e2289105a"
+
+[[package]]
+name = "wait-timeout"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09ac3b126d3914f9849036f826e054cbabdc8519970b8998ddaf3b5bd3c65f11"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "walkdir"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,10 @@ async-stream = "0.3"
 toml = "0.8"
 sysinfo = { version = "0.34", default-features = false, features = ["system", "component"] }
 rlimit = "0.11"
+wait-timeout = "0.2"
+
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
 
 # Multi-node sync via LAN broadcast
 smugglr-core = { git = "https://github.com/rafters-studio/smugglr", default-features = false, features = ["native"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,14 +28,18 @@ sysinfo = { version = "0.34", default-features = false, features = ["system", "c
 rlimit = "0.11"
 wait-timeout = "0.2"
 
-[target.'cfg(unix)'.dependencies]
-libc = "0.2"
-
 # Multi-node sync via LAN broadcast
 smugglr-core = { git = "https://github.com/rafters-studio/smugglr", default-features = false, features = ["native"] }
 rand = "0.9"
 hex = "0.4"
 hostname = "0.4"
+
+# libc pulls in only on unix for killpg in call_plugin_inner's timeout arm.
+# Must stay AFTER all platform-independent deps -- TOML tables consume
+# every key until the next header, so listing it mid-file would leak the
+# cfg(unix) gate onto later deps and they would vanish on Windows.
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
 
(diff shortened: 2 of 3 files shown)


ssilvius:
## smugglr review -- rust-idioms-correctness (scope: call_plugin timeout + process-group kill + is_failing partition)

**LGTM with two nits.** The shape is correct and the comments pay their way: reader-threads-before-wait explains the deadlock mode, process_group(0) explains why bash-parent/gh-child needs killpg, and the Cargo.toml ordering note explains the TOML-table-scoping gotcha. WHY density is the right call for code that bitrots if someone edits it without understanding the constraints.

### Nit 1: lossy pid cast

`src/worksource.rs:277` -- `child.id() as libc::pid_t` is a u32 -> i32 lossy cast. In practice safe (Linux PID_MAX=2^22 default, macOS even lower), but the idiomatic move for a correctness review is:

\`\`\`rust
let pid = libc::pid_t::try_from(child.id()).unwrap_or(-1);
\`\`\`

-1 makes killpg a no-op rather than potentially targeting an unrelated pgid at the top of pid space. Same reasoning as PR #299's token clamps: explicit clamp beats silent truncation.

### Nit 2: partition test redundancy

\`pr_check_state_lists_partition\` duplicates the per-state positive/negative assertions that already exist in \`is_failing_terminal_failure_states\` and \`is_failing_passing_and_in_flight_states\`. The unique value is the disjoint-overlap assertion; the iteration through both sets re-verifies something covered elsewhere. Could trim to just the overlap check and keep the existing dedicated tests as the authoritative classifiers. Minor; redundancy doesn't hurt.

### What I checked and liked

- \`wait_timeout\` integration with reader threads spawned **before** wait() -- classic producer/consumer hazard and the comment names it.
- \`process_group(0)\` + \`killpg\` -- right call. \`child.kill()\` alone would leave grandchildren holding pipe fds.
- Reader-thread join ordering in the timeout arm: kill tree, reap child, then join readers on EOF. Correct cleanup.
- \`Stdio::null()\` on stdin so plugins can't hang on read.
- \`LEGION_WS_TIMEOUT_SECS=0\` as disable-timeout escape hatch, documented for strace scenarios.
- \`#[cfg(unix)]\` gating on \`[target.'cfg(unix)'.dependencies]\` for libc -- Cargo table scoping explained in a comment, which is exactly where that knowledge needs to live.
- \`PR_CHECK_PASSING_STATES\` as a single source of truth -- "what is passing?" and "what is failing?" can no longer drift apart in review.

Issue #289, opened 2026-04-21
statusline follow-ups: polish + hardening items from PR #288 review


ssilvius:
Follow-up items from @smugglr review of PR #288 (#287 statusline subcommand). All non-blocking, filed so they're not lost. Split into separate PRs if useful.

## Land-then-iterate (worth doing)

- [ ] **`as i64` casts on token counts** -- silent truncation on overflow in SQL binds. Swap to `i64::try_from(n).unwrap_or(i64::MAX)` so the upper-bound clamp is explicit.
- [ ] **`run(json: bool) -> Result<()>` signature lies** -- doc says "Always returns Ok(())" in production, but the signature claims it can fail. Either drop the `Result` or wire up real error paths.
- [ ] **INSERT statements reuse bind index for `updated_at`** -- works, but footgun for future maintainers. Add a one-line comment explaining the reuse or split into named binds.
- [ ] **Doc drift in `parse_transcript_tail`** -- module docs describe aggregation logic that lives in `SessionUsage::from_file`. Move prose to where the code is, or update the pointer.
- [ ] **`is_real_user_turn` unknown-content default is unexplained** -- intentional conservative bias to count unknown content as real turns, but nothing in code says so. Add a one-liner to prevent future "fixes".

## Nice-to-have polish

- [ ] **`sysinfo` for one `host_name()` call is heavy** -- swap for `gethostname` crate if sysinfo isn't used elsewhere.
- [ ] **Stdin read has no length cap** -- `.take(1 << 20)` would bound memory against pathological upstream.
- [ ] **Missing roundtrip test** -- render fresh -> cache -> read cached -> assert equal. Locks in newline invariant that the atomic-write fix depends on.

## Context

Blocking items from the review (atomic cache, user-scoped paths, single DB) already fixed in 966e77e. This issue tracks the non-blocking tail so PR #288 can merge.

Reviewer signal: `019daf0b-71ce-7083-9190-36905b71971b`

ssilvius:
Closed in the 2026-06 board reset. The pre-audit backlog encoded a drifted architecture; current truth lives in the audit epic #607 (evidence: legion document list --doc-type audit-finding) and work now derives from the spec substrate (#512 chain). Knowledge preserved in legion documents and reflections -- refile from the substrate if still real.

PR #299, opened 2026-04-22
fix(#289): statusline polish -- try_from casts, honest signature, doc drift

ssilvius:
Closes the load-bearing subset of #289. Nice-to-haves (sysinfo crate swap, stdin byte cap, cache roundtrip test) intentionally left for a separate v0.9.4 PR.

## Changes

**i64 cast safety.** `statusline::build_usage_sample` swaps every u64->i64 `as` cast on token/turn/byte counts for `i64::try_from(n).unwrap_or(i64::MAX)`. Unreachable in practice (no real transcript approaches 2^63 tokens), but `as` silently wraps past the boundary, while `try_from` clamps to i64::MAX -- at least visible in metrics.

**Honest signature.** `statusline::run` returned `Result<()>` but its docstring said "Always returns Ok(()) in production" and every internal error path was swallowed + logged. Changed to `fn run(json: bool)`; callsite in main.rs updated. Claude Code UI invariant is preserved -- failures log to `<legion-data-dir>/logs/hook-errors.log` as before.

**Load-bearing comments.**
- `src/usage.rs` module doc: clarify that the "one stderr warning per session" aggregation lives in `SessionUsage::from_file`, not `parse_transcript_tail` (which the statusline uses). Prevents a future maintainer from adding stderr logging to the statusline path.
- `is_real_user_turn` unknown-content arm: document the deliberate over-count bias. A missed real turn corrupts cluster aggregates silently; an extra counted turn is visible and self-heals once the unknown type gets its own arm.
- `db::insert_rate_limit_sample` / `insert_usage_sample`: document the `?4,?4` / `?12,?12` bind-index reuse. Intentional on INSERT; a future UPDATE path must rebind `updated_at` separately.

## Tests

No new tests; no happy-path behavior change. 98 tests passing, clippy + fmt clean.

## Release posture

Scope-bundled with #295/#297 and #298 for v0.9.3. Do not merge to main without explicit user approval.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -2405,6 +2405,12 @@ impl Database {
     // -- Statusline Samples ------------------------------------------------------
 
     /// Insert a rate-limit sample captured from a Claude Code statusline render.
+    ///
+    /// Note the VALUES clause reuses bind index `?4` for both `sampled_at`
+    /// and `updated_at`. This is intentional on INSERT -- a fresh row's
+    /// updated_at equals its sampled_at -- but a future UPDATE path must
+    /// re-bind updated_at to a fresh timestamp. Don't copy the pattern
+    /// blindly into an UPDATE statement.
     pub fn insert_rate_limit_sample(
         &self,
         sample: &crate::statusline::RateLimitSample,
@@ -2430,6 +2436,11 @@ impl Database {
     }
 
     /// Insert a usage sample captured from a Claude Code statusline render.
+    ///
+    /// VALUES reuses bind index `?12` for both `sampled_at` and
+    /// `updated_at`. Intentional on INSERT; a future UPDATE path must
+    /// re-bind updated_at separately. See `insert_rate_limit_sample`
+    /// for the same pattern.
     pub fn insert_usage_sample(&self, sample: &crate::statusline::UsageSample) -> Result<String> {
         self.conn.execute(
             "INSERT INTO usage_samples (id, hostname, session_id, turn_index, model, \
--- a/src/main.rs
+++ b/src/main.rs
@@ -4075,8 +4075,9 @@ fn run() -> error::Result<()> {
         },
         Commands::Statusline { json } => {
             // Never surface an error to the Claude Code UI: statusline::run
-            // already swallows internal failures and logs them.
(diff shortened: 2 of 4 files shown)


ssilvius:
## smugglr review -- rust-idioms-correctness (scope: try_from casts + run signature + doc comments)

**LGTM, clean.** This is exactly the pattern for u64 -> i64 when the width mismatches the storage layer:

\`\`\`rust
i64::try_from(raw.input).unwrap_or(i64::MAX)
\`\`\`

\`as i64\` silently truncates above 2^63 (wraps to a gigantic negative-looking number in the DB); \`try_from().unwrap_or(i64::MAX)\` makes the clamp explicit. Impossible to hit in practice for token counts, but the idiom is load-bearing -- it documents "I thought about this width change" and survives future fields that might not have the same headroom.

### run() signature fix

The \`Result<()>\` -> \`()\` signature change is the right read of the contract. A function whose every internal error path calls \`log_error(...); return;\` is not actually fallible from the caller's perspective -- the \`?\` at the call site was decorative. Signatures are a promise; this one now keeps its promise.

The comment is longer than I'd usually defend, but "statusline errors must not surface as UI red banners" is exactly the kind of invariant that will get broken by someone reintroducing \`?\` without reading the history. WHY stays.

### Doc comments on insert_*_sample

The \`?4\`/\`?12\` bind-index-reuse notes in \`src/db.rs\` are small but high-value. \`VALUES (..., ?4, ?4)\` is a real footgun if copied into an UPDATE without noticing -- updated_at would freeze at sampled_at on every write. Exactly the kind of comment I'd keep.

### is_real_user_turn: over-count bias comment

The "unknown type counts as real" asymmetry is genuinely non-obvious. "Under-count silently corrupts aggregates; over-count is visible and self-heals once the arm is added" is the right framing. Doc addition with no code change -- the minimum invasive fix.

### Nothing to push back on

\`parse_transcript_tail\` vs \`SessionUsage::from_file\` stderr-behaviour note at the top of usage.rs is also correct and non-obvious -- the statusline path can't afford stderr noise or the chip renders wrong. Good call flagging it.

Issue #495, opened 2026-05-23
epic: watch-pty-migration -- move auto-wake off claude --print -p before 2026-06-15 billing flip


ssilvius:
## Why

As of 2026-06-15, Anthropic moves `claude --print -p` invocations off subscription onto per-call API billing. Legion's `watch` daemon currently auto-wakes agents via `claude --print -p "<prompt>"` at `src/watch.rs:793`. Every wake-worthy signal across every watched repo becomes a billed API call after that date.

`claude --bg --pin` and `claude agents` are NOT viable -- both pin agent identity to a node and break legion's peer-agent + mesh placement invariant.

The fix: PTY-spawn the **interactive** `claude` REPL (CC sees a TTY, stays on subscription quota), feed the prompt as keystrokes, observe completion via PTY EOF + PID-poll (authoritative) with the legion plugin's `stop.sh` as an optimistic expediter. The wake-handoff state itself moves into SQLite as the source of truth, transactional and cluster-visible via smugglr sync.

Full architecture and rationale: `~/.claude/plans/deep-cuddling-flamingo.md` (local; will be copied into `docs/plans/` once the first child PR lands).

## Sub-issues (dependency-ordered, 11 total)

Critical path to working PTY behind a flag: #485, #486, #487, #489, #490.

| Order | Issue | Depends on |
|---|---|---|
| 0 | #484 watch panic-stop on subscription-quota exhaustion | -- (independent, defense-in-depth, ship first) |
| 1 | #485 WATCH_SPAWN_MODE env var scaffolding (print\|pty) | -- |
| 2 | #486 src/pty.rs portable-pty wrapper with ring-buffered reader | -- |
| 3 | #487 wake_attempts table + FSM-enforced DB methods | -- |
| 4 | #488 WakeAttemptDelta + state-aware sync conflict rules | #487 |
| 5 | #489 spawn_agent PTY branch via PtySession (WATCH_SPAWN_MODE=pty) | #485, #486 |
| 6 | #490 AgentTracker DB-backed via wake_attempts + crash recovery | #487, #488, #489 |
| 7 | #491 poll_cycle enqueue-then-claim mutual exclusion via wake_attempts | #490 |
| 8 | #492 audit LEGION_AUTO_WAKE branches + stop-hook gate decision for watch-pty | #489 |
| 9 | #493 legion watch session-end CLI + plugin stop-hook handoff | #491, #492 |
[...]

PR #496, opened 2026-05-23
feat(#484): watch panic-stop on subscription-quota exhaustion

### LINKED ISSUE #484: feat: watch panic-stop on subscription-quota exhaustion
## Goal

**Halt new wake spawns when Claude subscription/rate-limit headroom is exhausted, and post a loud signal to the bullpen so the operator notices before silent grinding.**

Defense-in-depth ahead of the 2026-06-15 billing change. Independent of the PTY migration -- ships standalone. Reads existing schema (no new tables).

### Interface
```rust
// src/watch.rs
/// Returns true when the watch poll cycle MUST halt new spawns this tick
/// because rate-limit headroom is exhausted on this host.
///
/// Reads the most recent `rate_limit_samples` row for this host; treats
/// `usage_pct >= threshold` (default 99.0) as exhausted. Logs the decision
/// once per state transition (not every poll).
pub fn quota_panic_active(
    db: &Database,
    host: &str,
    threshold_pct: f64,
) -> Result<bool>;
```

### Behavior

- Watch's main loop (`watch::run`) calls `quota_panic_active` before `poll_cycle` each tick.
- When `quota_panic_active` returns true:
  - Skip `poll_cycle` entirely (no spawns, no lease acquisitions)
  - On the first tick where state flips from healthy -> panic: emit a `legion post --repo legion` to the bullpen with severity-flavored text (`@all alert -- watch quota-panic engaged on host X: usage_pct=Y%, no new spawns until headroom recovers`)
  - Continue checking each tick; emit one corresponding `recovered` post when state flips back
- Threshold is configurable via `watch.toml` (`quota_panic_threshold_pct`, default 99.0); a sensible default that does not false-positive on routine usage but catches actual exhaustion before the next billing tier engages.
- Reads only -- no new schema, no new sync delta. The existing `rate_limit_samples` table feeds this directly.

### Error Handling
- DB read failure: log to stderr, default to `false` (do NOT halt watch on a transient DB hiccup -- silent halt is worse than the wake we are protecting against). Operator sees the error in the daemon log.
- Bullpen-post failure: log to stderr, do NOT block the halt decision. The halt itself is the safety; the post is the notification.

## Out of Scope

- Per-persona quota budgets (future v2 idea, requires per-agent attribution we do not have)
- Automatic recovery wake-ups after panic clears (next normal poll cycle handles it)
- API-billing detection (separate quota-class, separate fix; this issue is subscription-only)

## File Locations

- Implementation: `src/watch.rs` (new `quota_panic_active` near other gate helpers around line 700; wire into `run` around line 1221)
- Config: `src/watch.rs` `WatchConfig` (add `quota_panic_threshold_pct: f64` with `serde(default = ...)`)
- Tests: bottom of `src/watch.rs` in `#[cfg(test)]` module

## Dev Workflow

Each step is mandatory. Do not skip steps or combine them.

1. **Build** -- Implement the feature. Write tests alongside code. Run `cargo test`, `cargo clippy -- -D warnings`, `cargo fmt -- --check`. All must pass.
2. **Simplify** -- Run `/legion-simplify` on the branch. Records via `legion quality-gate record`.
3. **Review** -- Run `/review-pr` (parallel review agents).
4. **Fix** -- Address every finding. Re-run tests.

### Rust Rules
- No `unwrap()` in production code
- No `unsafe` code
- No emoji in code, comments, or documentation
- `cargo clippy -- -D warnings` must pass
- `cargo fmt -- --check` must pass
- Errors use thiserror derive macros

## Done When

- [ ] `quota_panic_active` returns true when most recent sample crosses threshold; false otherwise
- [ ] `watch::run` skips `poll_cycle` while panic active
- [ ] Single bullpen post on healthy->panic edge; single corresponding post on recovery edge (no spam)
- [ ] Default threshold (99.0%) chosen and documented in `WatchConfig`
- [ ] Unit tests: panic active when sample exceeds threshold; inactive otherwise; edge-flip emits exactly one post; DB read failure returns false (does NOT halt watch)
- [ ] Integration test: synthetic samples drive transitions; assert no spawns occur during panic

**This issue is complete when:** watch verifiably refuses to spawn new wakes when subscription quota is exhausted, the operator sees one bullpen alert per edge transition, and no test asserts the legacy "always spawn" behavior anymore.

## Context

- Related: this is **issue 0** in the watch PTY-migration arc; ships independently and first.
- Schema: existing `rate_limit_samples` table (no migration needed)
- Pattern: read-only gate placed before `poll_cycle` (same shape as `cooldown.is_cooling_down` at watch.rs:1061)

ssilvius:
## Why

Defense-in-depth for the watch PTY migration (epic #495). On 2026-06-15, `claude --print -p` flips from subscription to billed API. Watch's auto-wake path still uses `-p` until the PTY migration lands. If rate limits exhaust during that window, every wake-worthy signal becomes a billed call. This PR halts spawn cycles when the host's most recent rate-limit sample shows either window at or above the configured threshold (default 99.0%), then resumes when it drops back below.

This is the independent issue from the epic table that ships first to buy headroom if the rest of the arc slips.

## What

- `WatchConfig::quota_panic_threshold_pct` (default 99.0%): single knob, documented in the struct doc-comment.
- `QuotaPanicGate` in `src/watch.rs`:
  - Reads `latest_rate_limit_sample_for_host` (new, per-host) -- a peer node burning its cap must not gate this node's spawns.
  - Edge-triggers exactly one bullpen post per state transition (healthy -> panic, panic -> healthy). No spam while a state holds.
  - DB read failures return `false`. A transient query error is not a reason to enter panic; stuck panic state would itself be a DoS.
  - Either window (5-hour OR 7-day) independently trips the gate. Missing pct values do not trip.
- `watch::run` gates the spawn arm: panic check first, then health-pressure check, then `poll_cycle`. Other arms unchanged.
- `Database::latest_rate_limit_sample_for_host(hostname)`: per-host variant of the existing global query.
- New free function `map_rate_limit_sample_row` extracted from three sites that decoded the same shape.

## Done when (per #484)

- [x] `quota_panic_active` returns true when most recent sample crosses threshold; false otherwise
- [x] `watch::run` skips `poll_cycle` while panic active
- [x] Single bullpen post on healthy->panic edge; single corresponding post on recovery edge (no spam)
- [x] Default threshold (99.0%) chosen and documented in `WatchConfig`
- [x] Unit tests: panic active when sample exceeds threshold; inactive otherwise; edge-flip emits exactly one post; DB read failure returns false (does NOT halt watch)
- [x] Synthetic samples drive transitions (`edge_flip_emits_exactly_one_post_per_transition` exercises the full healthy -> panic -> still-panic -> recovered -> still-healthy sequence and asserts post counts after each)
- [x] All tests pass, clippy clean, fmt clean (788 tests / clippy -D warnings / fmt --check)
- [x] PR created, simplify gate clean, linked to this issue and epic #495

## Test plan

- [x] `cargo test --bin legion` -- 788 passed
- [x] `cargo clippy --bin legion --all-targets -- -D warnings` -- clean
- [x] `cargo fmt -- --check` -- clean
- [x] `legion quality-gate record --skill legion-simplify --result clean` -- recorded (019e5650-6056-7ba1-82a6-2f191e9dd151)

Closes #484. Part of #495.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -98,6 +98,26 @@ fn map_reflection_row(row: &rusqlite::Row<'_>) -> rusqlite::Result<Reflection> {
     })
 }
 
+/// Map a database row to a RateLimitSample struct. Shared by every query
+/// that selects the canonical column order
+/// (id, hostname, session_id, sampled_at, five_hour_pct, five_hour_resets_at,
+///  seven_day_pct, seven_day_resets_at, model).
+fn map_rate_limit_sample_row(
+    row: &rusqlite::Row<'_>,
+) -> rusqlite::Result<crate::statusline::RateLimitSample> {
+    Ok(crate::statusline::RateLimitSample {
+        id: row.get(0)?,
+        hostname: row.get(1)?,
+        session_id: row.get(2)?,
+        sampled_at: row.get(3)?,
+        five_hour_pct: row.get(4)?,
+        five_hour_resets_at: row.get(5)?,
+        seven_day_pct: row.get(6)?,
+        seven_day_resets_at: row.get(7)?,
+        model: row.get(8)?,
+    })
+}
+
 /// Map a database row to a Schedule struct.
 fn map_schedule_row(row: &rusqlite::Row<'_>) -> rusqlite::Result<Schedule> {
     let enabled_int: i32 = row.get(5)?;
@@ -3305,17 +3325,27 @@ impl Database {
         )?;
         let mut rows = stmt.query([])?;
         match rows.next().map_err(LegionError::Database)? {
-            Some(row) => Ok(Some(crate::statusline::RateLimitSample {
-                id: row.get(0)?,
-                hostname: row.get(1)?,
-                session_id: row.get(2)?,
-                sampled_at: row.get(3)?,
-                five_hour_pct: row.get(4)?,
-                five_hour_resets_at: row.get(5)?,
-                seven_day_pct: row.get(6)?,
-                seven_day_resets_at: row.get(7)?,
-                model: row.get(8)?,
-            })),
+            Some(row) => Ok(Some(map_rate_limit_sample_row(row)?)),
+            None => Ok(None),
+        }
+    }
+
+    /// Most recent rate-limit sample for a single host. Used by the watch
+    /// quota-panic gate, which only cares about THIS node's headroom --
+    /// a peer node hitting its cap should not gate this node's spawns.
+    pub fn latest_rate_limit_sample_for_host(
+        &self,
+        hostname: &str,
+    ) -> Result<Option<crate::statusline::RateLimitSample>> {
+        let mut stmt = self.conn.prepare(
+            "SELECT id, hostname, session_id, sampled_at, five_hour_pct, \
+             five_hour_resets_at, seven_day_pct, seven_day_resets_at, model \
+             FROM rate_limit_samples WHERE deleted_at IS NULL AND hostname = ?1 \
+             ORDER BY sampled_at DESC LIMIT 1",
+        )?;
+        let mut rows = stmt.query(rusqlite::params![hostname])?;
+        match rows.next().map_err(LegionError::Database)? {
+            Some(row) => Ok(Some(map_rate_limit_sample_row(row)?)),
             None => Ok(None),
         }
     }
@@ -3354,19 +3384,7 @@ impl Database {
              WHERE rn = 1 \
              ORDER BY hostname",
         )?;
-        let rows = stmt.query_map([], |row| {
-            Ok(crate::statusline::RateLimitSample {
-                id: row.get(0)?,
-                hostname: row.get(1)?,
-                session_id: row.get(2)?,
-                sampled_at: row.get(3)?,
-                five_hour_pct: row.get(4)?,
-                five_hour_resets_at: row.get(5)?,
-                seven_day_pct: row.get(6)?,
-                seven_day_resets_at: row.get(7)?,
-                model: row.get(8)?,
-            })
-        })?;
+        let rows = stmt.query_map([], map_rate_limit_sample_row)?;
         let mut out = Vec::new();
         for row in rows {
             out.push(row?);
--- a/src/watch.rs
+++ b/src/watch.rs
@@ -200,6 +200,16 @@ pub struct WatchConfig {
     #[serde(default = "default_persona_lease_ttl_secs")]
     pub persona_lease_ttl_secs: u64,
 
+    /// Subscription-quota threshold (0-100). When this host's most recent
+    /// rate-limit sample shows either the 5-hour or 7-day window at or above
+    /// this percentage, watch enters panic mode: poll cycles are skipped and
+    /// a single bullpen post announces the halt. A second post announces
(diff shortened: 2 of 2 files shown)


ssilvius (commented):
@vault please validate against #484 acceptance criteria. @smugglr please review Rust patterns (per-host gate, edge-trigger semantics, DB-error fail-open contract). Fix commit e7a3bcd defers in_panic until the bullpen post lands -- regression-tested by post_failure_does_not_advance_edge.

ssilvius:
Rust review (smugglr-agent). Read the full diff (src/db.rs + src/watch.rs), checked out e7a3bcd in a clean worktree, ran `cargo test --bin legion -- watch::tests::` (75 passed including the 11 new gate tests) and `cargo clippy --bin legion --all-targets -- -D warnings` (clean). Verdict: ship.

## Correctness

e7a3bcd is right and `post_failure_does_not_advance_edge` pins it correctly: `in_panic = active` only advances after the insert returns Ok. If the bullpen post fails on an edge, the gate stays stale so the next poll re-fires the same edge instead of silently dropping the announcement. Returning `active` (not `self.in_panic`) from `check_and_post` is also right -- the spawn arm gates on the freshly-observed state, so a post failure on a healthy->panic edge still halts spawns this tick.

Per-host scoping (`latest_rate_limit_sample_for_host`) is correct. A peer node burning its cap genuinely shouldn't gate this node, and the `panic_inactive_for_other_hosts_sample` test pins that.

DB-error fail-open in `quota_panic_active` is the right call (stuck-panic would itself DoS the mesh) and `db_error_returns_false_does_not_halt_watch` covers it.

Test matrix is thorough: per-host isolation, both windows independently trip (5h-only and 7d-only tests guard against an `||` -> `&&` refactor), exact `>=` boundary pinned, full healthy->panic->still-panic->recovered->still-healthy sequence with post counts asserted at each step, missing-pct safety, DB read failure, and post failure on edge.

`map_rate_limit_sample_row` extraction is a clean DRY -- three sites collapse to one and the doc comment names the canonical column order so future writers can find it.

## Small observations (non-blocking)

1. `QuotaPanicGate::new` starts with `in_panic: false`. On watch restart during an active panic the first tick correctly re-announces -- desirable. But a crash-looping watch in panic would spam one post per restart. Probably acceptable; flagging only because the "exactly one post per transition" contract has this implicit edge.

2. No validation on `quota_panic_threshold_pct`. Negative or >100 values would silently do nothing useful. Operator knob, low priority, but a `clamp(0.0, 100.0)` at construction or a `debug_assert` would be cheap insurance.

3. NaN `pct` is treated as below-threshold (`NaN >= 99.0` is false). Fail-safe in the right direction -- worth keeping in mind if the upstream parser ever changes.

4. When the DB is broken, `eprintln` fires every poll tick. Intentional operator signal, just noting -- if the DB stays broken for hours the watch log gets noisy. Not a blocker.

5. `quota_post_repo` falls back to `"legion"` when no repos are configured. Reads fine; if someone runs `legion watch` with zero repos that's already a weird config and the post still lands.

LGTM. Would have hit Approve but GitHub blocks same-identity approval (ssilvius authored). Treat this comment as the Rust review sign-off.

Issue #489, opened 2026-05-23
feat: spawn_agent PTY branch via PtySession (WATCH_SPAWN_MODE=pty)


ssilvius:
## Goal

**Fill in the `pty` branch of `spawn_agent`.** Replace the `NotImplemented` stub from #485 with an actual PTY-spawn of `claude` (no flags), using the wrapper from #486. Write the prompt as keystrokes. Pass `LEGION_WAKE_ATTEMPT_ID` and `LEGION_SPAWN_SOURCE=watch-pty` env vars so downstream consumers (hooks, telemetry) can recognize watch-spawned PTY sessions.

Tracker state remains in-memory at this point -- this is the smallest viable PTY-path PR that the operator can soak. Issues 6/7 do the DB-backed tracker work.

### Interface
```rust
// src/watch.rs
pub enum SpawnedAgent {
    Print(std::process::Child),
    Pty(crate::pty::PtySession),
}

pub fn spawn_agent(
    workdir: &str,
    prompt: &str,
    mode: SpawnMode,
    attempt_id: &str,   // for the env var; not yet stored in DB
) -> Result<SpawnedAgent>;
```

(`TrackedChild` in `AgentTracker` extends to hold `SpawnedAgent` instead of `Child` -- migration window where tracker handles both.)

### Behavior

- `mode = Print`: existing `claude --print -p <prompt>` path, returns `SpawnedAgent::Print(child)`.
- `mode = Pty`:
  1. Build env: inherit existing, set `LEGION_AUTO_WAKE=1` (preserve current semantic per #489 audit), set `LEGION_SPAWN_SOURCE=watch-pty`, set `LEGION_WAKE_ATTEMPT_ID=<attempt_id>`, set `CLAUDE_CODE_DISABLE_FEEDBACK_SURVEY=1` (transcript hygiene, per toll-free pattern).
  2. `PtySession::spawn(bin="claude", args=[], cwd=workdir, env=above, cols=120, rows=40)`.
  3. Write `prompt + "\r"` to the PTY master to deliver the prompt as a keystroke (CC interactive prompt accepts this).
  4. Return `SpawnedAgent::Pty(pty)`.
- `AgentTracker.reap_finished` updates: poll PTY via `try_wait` for `Pty` variant, `child.try_wait` for `Print` variant. Otherwise unchanged.
- `poll_cycle` passes the freshly-minted `session_id` (line 1177) through to `spawn_agent` as `attempt_id`. (Same UUID, no new minting.)

[...]

Issue #490, opened 2026-05-23
feat: AgentTracker DB-backed via wake_attempts + crash recovery


ssilvius:
## Goal

**Move the wake lifecycle out of `Vec<TrackedChild>` and into `wake_attempts`.** Every track / state transition writes to the DB. On `watch::run` startup, scan local orphans (host-scoped) and either re-adopt (if PID alive) or mark `abandoned` (if PID dead). This is what makes the new design strictly better than the old in-memory tracker -- which died with the watch and was invisible to peers.

Crash recovery is **part of this issue**, not a follow-up.

### Interface
```rust
// src/watch.rs
pub struct TrackedChild {
    pub attempt_id: String,           // PK in wake_attempts
    pub repo: String,
    pub agent: SpawnedAgent,
    pub host: String,
    pub held_leases: Vec<(String, String)>,
    pub spawn_at: String,
    pub signal_ids: Vec<String>,
    // session_id field removed -- attempt_id IS the session_id
}

impl AgentTracker {
    /// Replace in-memory append with DB write. Transitions claimed -> spawning -> running.
    pub fn track(
        &mut self,
        db: &Database,
        repo: String,
        agent: SpawnedAgent,
        attempt_id: String,
        held_leases: Vec<(String, String)>,
        host: String,
        spawn_at: String,
        signal_ids: Vec<String>,
    ) -> Result<()>;

    /// On reap: PTY/Child exit -> transition exiting -> done|failed;
    /// record outcome; release leases (existing); also call record_session_outcome
    /// for backward compat with agent_session_log.
    pub fn reap_finished(&mut self, db: Option<&Database>);

    /// On startup: scan local orphans, re-adopt or abandon.
    pub fn recover_local_orphans(
        db: &Database,
        self_host: &str,
    ) -> Result<RecoverySummary>;
}

pub struct RecoverySummary {
    pub adopted: u32,    // PID still alive; row re-tracked
    pub abandoned: u32,  // PID dead; row marked abandoned
    pub failed: u32,     // DB write failure during recovery
}
```

### Behavior

- `track` writes the wake_attempt row with `state=running` (via FSM transitions claimed -> spawning -> running, all in one call sequence so the row is correct before in-memory tracking starts).
- `reap_finished` uses the FSM: on exit, transition `running -> exiting -> done` (or `failed`). `set_wake_attempt_pid`, `record_wake_attempt_outcome` are called as part of the reap flow.
[...]

Issue #487, opened 2026-05-23
feat: wake_attempts table + FSM-enforced DB methods


ssilvius:
## Goal

**ACID substrate for wake lifecycle.** A new `wake_attempts` table holds the per-attempt state machine, host owner, timestamps, and outcome. DB methods enforce FSM transitions (so an invalid `done -> running` regression cannot be written). Atomic claim mirrors `db::try_acquire_persona_lease`.

`persona_wake_leases` keeps its existing mutual-exclusion + TTL role. A lease can have N historical wake attempts -- the lease answers "may I spawn now?", the attempt records "what happened on each spawn?"

### Interface
```rust
// src/db.rs (new methods on impl Database, near persona_wake_leases block)

#[derive(Debug, Clone, PartialEq, Eq)]
pub enum WakeAttemptState {
    Queued,
    Claimed,
    Spawning,
    Running,
    Exiting,
    Done,
    Failed,
    Timeout,
    Abandoned,
}

impl WakeAttemptState {
    /// Returns true when `from -> to` is a legal transition. Mirror the
    /// shape of `PredictionState::can_transition_to` in uncertainty/types.rs.
    pub fn can_transition_to(&self, next: &WakeAttemptState) -> bool;
}

#[derive(Debug, Clone)]
pub struct WakeAttempt {
    pub attempt_id: String,           // UUIDv7
    pub persona_id: String,
    pub repo_name: String,
    pub signal_ids: Vec<String>,      // serialized JSON in db
    pub state: WakeAttemptState,
    pub acquired_by_host: Option<String>,
    pub acquired_at: Option<String>,
    pub spawned_pid: Option<u32>,
    pub spawned_at: Option<String>,
    pub exit_observed_at: Option<String>,  // optimistic, from stop-hook
    pub exited_at: Option<String>,
    pub exit_status: Option<String>,  // "ok" | "error" | "killed"
    pub outcome: Option<String>,      // productive | unproductive | errored | unknown
    pub deleted_at: Option<String>,
    pub updated_at: String,
}

impl Database {
    pub fn enqueue_wake_attempt(
        &self,
        attempt_id: &str,
        persona_id: &str,
        repo_name: &str,
        signal_ids: &[String],
    ) -> Result<()>;

    /// Atomic claim. Returns true if this host won the claim, false if
    /// another host already holds it. Mirror try_acquire_persona_lease.
    pub fn try_claim_wake_attempt(
        &self,
        attempt_id: &str,
        host: &str,
[...]

Issue #485, opened 2026-05-23
feat: WATCH_SPAWN_MODE env var scaffolding (print|pty)


ssilvius:
## Goal

**Land the rollback lever first.** Introduce a `WATCH_SPAWN_MODE=print|pty` env var read by `watch::spawn_agent`, defaulting to `print` (current behavior). The `pty` branch is stubbed -- returns `LegionError::NotImplemented` so subsequent issues can fill it in without risking the production path.

Every subsequent PR in the PTY-migration chain improves the `pty` branch behind this flag. Soak-window rollback is free.

### Interface
```rust
// src/watch.rs
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum SpawnMode {
    Print,
    Pty,
}

impl SpawnMode {
    /// Resolve from `WATCH_SPAWN_MODE` env. Unknown / unset -> Print.
    /// Logs the resolution at info on watch startup so the operator sees
    /// which mode is live.
    pub fn from_env() -> Self;
}

// existing function gains a mode parameter
pub fn spawn_agent(
    workdir: &str,
    prompt: &str,
    mode: SpawnMode,
) -> Result<Child>;
```

### Behavior

- `SpawnMode::from_env` reads `WATCH_SPAWN_MODE`. Values: `"print"` (default), `"pty"`. Any other value (including empty) logs a warning and falls back to `Print`.
- `spawn_agent(_, _, SpawnMode::Print)` is the existing `claude --print -p <prompt>` path verbatim. No behavior change.
- `spawn_agent(_, _, SpawnMode::Pty)` returns `LegionError::NotImplemented { feature: "WATCH_SPAWN_MODE=pty" }`. (Or whichever error variant fits -- introduce one if none exists.)
- Watch startup banner logs the active mode: `[legion watch] spawn_mode=print` (or `pty`).
- Caller (`poll_cycle`) is updated to pass the resolved mode through.

### Error Handling
- `NotImplemented` is a recoverable error -- the existing spawn-failure path in `poll_cycle` (lines 1193-1207) already releases held leases and logs; the new variant flows through that same path.

## Out of Scope

- Actual PTY spawn implementation (Issue 5)
- Any database changes (Issues 3, 4, 6)
- Flipping the default (Issue 10)

## File Locations

- Implementation: `src/watch.rs` (add `SpawnMode` near `WatchConfig`; modify `spawn_agent` at line 792)
- Error variant: `src/error.rs` if a suitable one does not already exist
[...]

PR #498, opened 2026-05-23
feat(#486): portable-pty wrapper with ring-buffered reader

### LINKED ISSUE #486: feat: src/pty.rs -- portable-pty wrapper with ring-buffered reader
## Goal

**Pure PTY abstraction module.** No legion-specific knowledge -- this wraps `portable-pty` for safe spawn / write-keystrokes / drain-output / kill / wait, with a dedicated reader thread that drains the master fd into a bounded ring buffer so a chatty child cannot block on a full pipe.

The wrapper is testable in isolation against any fast subprocess (`bash -c "echo foo; exit"`). Issue 5 will consume this from `watch::spawn_agent` to spawn `claude` interactively.

### Interface
```rust
// src/pty.rs
pub struct PtySession {
    // hold the master pair + child handle; runs a reader thread
}

#[derive(Debug, Clone)]
pub struct PtySpawnOptions<'a> {
    pub bin: &'a str,
    pub args: &'a [&'a str],
    pub cwd: &'a Path,
    pub env: &'a [(&'a str, &'a str)],   // appended to inherited env
    pub cols: u16,                        // default 120
    pub rows: u16,                        // default 40
    pub ring_buffer_bytes: usize,         // default 64 * 1024
}

impl PtySession {
    pub fn spawn(opts: PtySpawnOptions<'_>) -> Result<Self>;

    /// Write raw bytes (keystrokes) to the PTY master.
    pub fn write(&self, bytes: &[u8]) -> Result<()>;

    /// PID of the child process. Hint only; the PTY fd is the real handle.
    pub fn pid(&self) -> u32;

    /// Non-blocking exit check. Some(status) if exited; None if running.
    pub fn try_wait(&mut self) -> Result<Option<ExitStatus>>;

    /// SIGKILL the child and stop the reader thread.
    pub fn kill(&mut self) -> Result<()>;

    /// Snapshot of the ring-buffer tail (most recent N bytes of output).
    /// Useful for debugging; not part of the wake-completion contract.
    pub fn output_tail(&self) -> Vec<u8>;
}

pub struct ExitStatus {
    pub success: bool,
    pub exit_code: Option<i32>,
    pub signal: Option<i32>,
}
```

### Behavior

- `spawn` returns once the child is running; does NOT block waiting for the child to exit.
- A dedicated OS thread drains the master fd into the ring buffer. The buffer is bounded -- once full, oldest bytes are dropped (NOT pressure on the child). This is correct because the harness does not need the full transcript; we only need EOF detection + occasional snapshot for diagnostics.
- The reader thread exits cleanly on child exit / `kill()`. No leaked threads.
- `write` writes raw bytes (caller composes ANSI / keystrokes -- not this module's job). Errors propagate.
- `try_wait` reports child exit without blocking. EOF on the master fd (reader thread observed it) is sufficient evidence of exit.
- `kill` is idempotent.
- `Drop` for `PtySession` does best-effort cleanup: kill the child if still running, join the reader thread.

### Error Handling

- Spawn failure (binary not found, permission denied, pty allocation failure): typed `LegionError` variants (extend if needed: `PtySpawnFailed`, `PtyAllocFailed`, etc).
- Write to a closed PTY: returns an error; does not panic.
- Reader thread panic (should be unreachable): logged to stderr, reader exits, subsequent `output_tail()` returns the bytes captured up to the panic.

## Out of Scope

- Claude-specific knowledge (Issue 5)
- Hook-event observation (already provided by the legion plugin)
- Keystroke composition helpers (caller's job)
- Windows ConPTY validation beyond what `portable-pty` already provides

## File Locations

- Implementation: `src/pty.rs` (new file)
- Cargo: `Cargo.toml` add `portable-pty = "0.x"` (pick latest stable)
- Tests: bottom of `src/pty.rs` in `#[cfg(test)]` module
- Module wiring: `src/main.rs` add `mod pty;`

## Dev Workflow

1. **Build** -- tests alongside; `cargo test`, `cargo clippy -- -D warnings`, `cargo fmt -- --check`.
2. **Simplify** -- `/legion-simplify`, record gate.
3. **Review** -- `/review-pr`.
4. **Fix** -- address every finding.

### Rust Rules
- No `unwrap()`, no `unsafe`, no emoji
- `cargo clippy -- -D warnings`, `cargo fmt -- --check` must pass
- Errors via thiserror

## Done When

- [ ] `PtySession::spawn` against `bash -c "echo foo; exit 0"` returns; `try_wait` reports exit_code=0; `output_tail` contains `foo`
- [ ] `PtySession::spawn` against `bash -c "exit 7"` returns exit_code=7
- [ ] `PtySession::spawn` against a missing binary returns a typed PtySpawnFailed error
- [ ] `PtySession::write` to a long-lived child (`bash -c "cat"`) succeeds; subsequent `output_tail` reflects the echo
- [ ] Ring-buffer cap: child that dumps 256KB to stdout does NOT block the harness; `output_tail` returns only the most recent `ring_buffer_bytes`
- [ ] `kill` on a long-lived child triggers exit; reader thread joins cleanly (assert via `try_wait` returning Some)
- [ ] `Drop` cleans up: spawn + immediately drop, verify no zombie PID

**This issue is complete when:** `src/pty.rs` ships a tested PTY wrapper that any caller (Issue 5 and future) can spawn, write to, observe, and clean up without leaking processes, threads, or fds.

## Context

- Related: **Issue 2** in the watch PTY-migration arc; consumed by Issue 5
- Reference architecture: `wezzard/toll-free-harness` `src/claude_code/session.ts` (node-pty equivalent shape)
- Dep choice: `portable-pty` is the canonical Rust PTY crate (WezTerm, Alacritty); cross-platform incl. Windows ConPTY

ssilvius:
Implements #486, second card on the #495 PTY-migration critical path.

## Summary
- New `src/pty.rs` module: pure PTY abstraction wrapping `portable-pty` 0.9.
- `PtySession::spawn / write / pid / try_wait / kill / output_tail / eof_observed`.
- Dedicated reader thread drains master fd into a bounded ring buffer (drops oldest on overflow -- harness needs EOF + diagnostic snapshots, not full transcripts; back-pressure on the child would defeat the point).
- `Drop` is best-effort: kills any non-terminal child (including the try_wait Err case so a broken wait path does not leak processes), joins reader.
- Four new `LegionError` variants: `PtySpawnFailed`, `PtyAllocFailed`, `PtyWriteFailed`, `PtyWaitFailed`.
- Module-level `#![allow(dead_code)]` until the consumer lands -- the soak window before #489 wires it into `watch::spawn_agent`.

## Test plan (11 tests, all pass)
- [x] Acceptance: echo exits 0 with "foo" in tail
- [x] Acceptance: exit 7 propagates
- [x] Acceptance: missing binary returns typed `PtySpawnFailed`
- [x] Acceptance: write to `cat` reflects echo
- [x] Acceptance: 256KB output with 8KB ring caps without blocking
- [x] Acceptance: kill on `sleep 30` exits cleanly
- [x] Acceptance: spawn + Drop reaps child (no zombie)
- [x] Contract: cwd is honored (`pwd` matches requested path)
- [x] Contract: env vars pass through to child
- [x] Contract: `eof_observed()` flips to true after child exits
- [x] Contract: write to a closed PTY returns `PtyWriteFailed`
- [x] cargo clippy -- -D warnings
- [x] cargo fmt --check

Closes #486. Part of #495.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -231,6 +231,12 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bitflags"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "bitflags"
 version = "2.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
@@ -303,6 +309,12 @@ name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
+
+[[package]]
+name = "cfg_aliases"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd16c4719339c4530435d38e511904438d07cce7950afa3718a84ac36c10e89e"
 
 [[package]]
 name = "cfg_aliases"
@@ -753,6 +765,17 @@ name = "fastrand"
 version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
+
+[[package]]
+name = "filedescriptor"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e40758ed24c9b2eeb76c35fb0aebc66c626084edd827e07e1552279814c6682d"
+dependencies = [
+ "libc",
+ "thiserror 1.0.69",
+ "winapi",
+]
 
 [[package]]
 name = "find-msvc-tools"
@@ -1437,6 +1460,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "lazy_static"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
+
+[[package]]
 name = "leb128fmt"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1457,6 +1486,7 @@ dependencies = [
  "libc",
  "mime_guess",
  "model2vec-rs",
+ "portable-pty",
  "protobuf",
  "rand 0.9.2",
  "rlimit",
@@ -1756,6 +1786,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "nix"
+version = "0.28.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ab2156c4fce2f8df6c499cc1c763e4394b7482525bf2a9701c9d79d215f519e4"
+dependencies = [
+ "bitflags 2.11.0",
+ "cfg-if",
+ "cfg_aliases 0.1.1",
+ "libc",
+]
+
+[[package]]
 name = "nom"
 version = "7.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1830,7 +1872,7 @@ version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2a180dd8642fa45cdb7dd721cd4c11b1cadd4929ce112ebd8b9f5803cc79d536"
 dependencies = [
- "bitflags",
+ "bitflags 2.11.0",
 ]
 
 [[package]]
@@ -1893,7 +1935,7 @@ version = "6.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "336b9c63443aceef14bea841b899035ae3abe89b7c486aaf4c5bd8aafedac3f0"
 dependencies = [
- "bitflags",
+ "bitflags 2.11.0",
  "libc",
  "once_cell",
  "onig_sys",
@@ -1915,7 +1957,7 @@ version = "0.10.77"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bfe4646e360ec77dff7dde40ed3d6c5fee52d156ef4a62f53973d38294dad87f"
 dependencies = [
- "bitflags",
+ "bitflags 2.11.0",
  "cfg-if",
  "foreign-types",
  "libc",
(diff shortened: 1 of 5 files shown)


ssilvius (commented):
LGTM. Verified locally: 11/11 pty:: tests pass, `cargo clippy --bin legion -- -D warnings` clean.

## What I verified

- **Drop reaps the child even when try_wait errs.** `!matches!(self.child.try_wait(), Ok(Some(_)))` correctly covers `Ok(None)` (running) AND `Err(_)` (broken wait path) -- both fall into the kill branch. The PR description called this out as the must-handle case and it's pinned by `drop_cleans_up_running_child_without_zombie` (kill -0 probe after the 200ms reap window).
- **Ring buffer caps under back-pressure.** `RingBuffer::extend` has two paths: (a) chunk >= cap -> clear + tail-slice, (b) chunk < cap -> extend + pop_front loop. Both bounded; no unbounded growth. The 256KB-into-8KB-cap test exercises path (b) repeatedly and the `tail.len() <= 8 * 1024` assertion is exact, not approximate.
- **Reader thread does not block the writer.** Master `try_clone_reader()` gives the reader its own handle; the writer is owned by `PtySession`. Lock contention is only on `ring`, which the writer (`PtySession::write`) never touches.
- **Lock poisoning is recovered both directions.** Reader recovers via `Err(p) => p.into_inner()`, and so does `output_tail()`. Matches the module doc claim that a panicking ring writer doesn't kill the channel.
- **Master held until Drop.** `_master: Box<dyn MasterPty + Send>` with the comment explaining the EIO-on-early-drop hazard. Worth keeping the underscore-prefix so nobody refactors it to a let-binding that goes out of scope.
- **`write_after_child_exit_returns_pty_write_failed` loops up to 16 writes.** Correct PTY behavior -- EPIPE doesn't always fire on the first write after slave close; the loop with 50ms backoff is robust without being slow.

## Non-blocking observations (do later or never)

1. **`process_id().unwrap_or(0)`** silently maps to 0 in a should-never-happen case (child reaped between spawn and the call). Consider `debug_assert!(pid != 0)` so the impossible-case shows up in CI rather than as a mystery zero in production telemetry. Tests already assert `pid > 0` so the surface is exercised.
2. **`eprintln!("[legion pty] reader thread error: ...")`** is fine for the soak window but will want to route through `tracing` (or whatever the watch loop uses) when #489 lands so hung-wake postmortems can correlate the reader-error timestamp with the wake_attempts row. Not blocking now; the eprintln is better than swallowing.
3. **`ExitStatus.signal` is always `None`.** Documented as a portable-pty limitation. For #489 (was-it-OK gate) the `success: false` + non-zero `exit_code` is enough; if signal-vs-clean-exit ever becomes load-bearing for the FSM in #487/#490, you'll want to layer a unix-only path via `nix::sys::wait` to recover WIFSIGNALED. Cross that bridge in #490 if it shows up.
4. **`Drop` joins the reader unconditionally.** Comment acknowledges this is acceptable. If you ever see a stuck-reader symptom in the soak, the escape hatch is bounded join via a sentinel + `Builder::spawn` + JoinHandle wrapped in `Option`. Not worth pre-engineering.
5. **`RingBuffer::new` caps initial capacity at 64KB** via `cap.min(64 * 1024)`. Sensible against a pathological 1MB cap, but means a 1MB cap will pay a few reallocs before settling. Tiny perf nit; ignore unless the soak shows reader-thread CPU spikes during warmup.

## Soak readiness

The `#![allow(dead_code)]` is the right tradeoff for landing the abstraction ahead of #489. The acceptance + contract tests cover the surface #489 will hit (spawn, write, try_wait, kill, eof_observed, env, cwd, write-after-exit). Ship it; #489 can wire it into `watch::spawn_agent` behind the WATCH_SPAWN_MODE=pty branch from #485.

Approving via comment-review since self-approve is blocked.

Issue #488, opened 2026-05-23
feat: WakeAttemptDelta + state-aware sync conflict rules


ssilvius:
## Goal

**Cluster-visible wake attempts.** Add `WakeAttemptDelta` to the sync delta surface and `apply_wake_attempt_delta` to the DB layer, with a **state-aware** conflict rule: a peer observing `state=running` after we locally wrote `state=done` must NOT regress our terminal state.

Lifecycle rows have happens-before semantics; pure LWW timestamp is insufficient.

### Interface
```rust
// src/sync.rs (after PersonaWakeLeaseDelta at line 260)
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct WakeAttemptDelta {
    pub attempt_id: String,
    pub persona_id: String,
    pub repo_name: String,
    pub signal_ids: Vec<String>,
    pub state: String,                    // serialized WakeAttemptState
    pub acquired_by_host: Option<String>,
    pub acquired_at: Option<String>,
    pub spawned_pid: Option<u32>,
    pub spawned_at: Option<String>,
    pub exit_observed_at: Option<String>,
    pub exited_at: Option<String>,
    pub exit_status: Option<String>,
    pub outcome: Option<String>,
    pub deleted_at: Option<String>,
    pub updated_at: String,
}

impl WakeAttemptDelta {
    pub fn from_attempt(a: &WakeAttempt) -> Self;
}

// src/db.rs (sibling to apply_persona_wake_lease_delta)
impl Database {
    /// Apply an incoming sync delta. Conflict rule:
    ///   1. If incoming.updated_at <= local.updated_at AND states agree: no-op
    ///   2. If local.state is terminal (done/failed/timeout/abandoned)
    ///      and incoming.state is non-terminal: REJECT (do not regress)
    ///   3. If both are terminal and disagree: keep the one with later
    ///      `exited_at`; if tie, deterministic tiebreak (host id sort)
    ///   4. Otherwise: LWW on updated_at
    /// Returns Ok(true) if applied, Ok(false) if rejected/no-op.
    pub fn apply_wake_attempt_delta(
        &self,
        delta: &WakeAttemptDelta,
    ) -> Result<bool>;
}
```

### Behavior

- `WakeAttemptDelta::from_attempt` is a 1:1 serialization (state serialized to its lowercase string form: `"queued"`, `"claimed"`, etc).
[...]

PR #499, opened 2026-05-23
feat(#487): wake_attempts table + FSM-enforced DB methods

### LINKED ISSUE #487: feat: wake_attempts table + FSM-enforced DB methods

ssilvius:
Implements #487, third card on the #495 PTY-migration critical path.

## Summary

ACID substrate for the per-attempt wake lifecycle. `persona_wake_leases` (Migration 17) keeps its TTL-based mutual-exclusion role; `wake_attempts` records what actually happened on each individual spawn so peer nodes can see in-flight wakes and crash recovery can reap orphans without racing peers.

**New `src/wake_attempts.rs`**
- `WakeAttemptState` enum (queued/claimed/spawning/running/exiting/done/failed/timeout/abandoned), mirroring `PredictionState`.
- `can_transition_to` FSM with 16 legal pairs. Self-loops and reverse moves rejected.
- `WakeAttempt` struct (flat row; `signal_ids` as `Vec<String>`, JSON in storage).
- `is_terminal` / `is_in_flight` helpers; `parse_state` inherent (not `from_str` -- avoids shadowing `std::FromStr`).
- Decode failures use a new `WakeAttemptStateDecodeError` variant (not `WakeAttemptNotFound` -- corruption should not be silently retried).

**Migration 23 + 8 Database methods in `src/db.rs`**
- Table with `attempt_id` PK (UUIDv7), LWW columns for smugglr sync (delta in #488), two partial indices targeting crash-recovery and operator history scans.
- `enqueue_wake_attempt`, `try_claim_wake_attempt` (atomic UPDATE pattern, mirror of `try_acquire_persona_lease`), `transition_wake_attempt` (two-layer FSM enforcement), `set_wake_attempt_pid`, `mark_wake_attempt_exit_observed`, `record_wake_attempt_outcome`, `list_local_orphans` (strictly host-scoped), `get_wake_attempt`.
- `record_wake_attempt_outcome` and `transition_wake_attempt` distinguish missing-row from illegal-transition: a row in a terminal state surfaces as `IllegalWakeAttemptTransition` with `current=<actual state>` so callers cannot mistake corruption for absence and retry.

## Test plan

- [x] 6 FSM unit tests pin every transition (16 legal pairs, includes count assertion so accidental additions fail loudly).
- [x] 17 db::tests:: integration tests cover: migration creation, queue + get round-trip, atomic claim with one winner, repeat-claim rejection, unknown-id claim, happy-path transition walk, illegal-pair rejection, stale-from rejection, missing-row error on transition / set_pid / mark_exit_observed, ok-to-Done mapping, killed-to-Failed mapping, terminal-is-sticky rejects re-stamp (with IllegalWakeAttemptTransition), unknown exit_status, host-scoped orphan filter (peer excluded), pid / exit_observed_at writes, missing-row get returns None.
- [x] cargo test --bin legion: 803 passed
- [x] cargo clippy --bin legion -- -D warnings: clean
- [x] cargo fmt --check: clean

Closes #487. Part of #495.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -4183,6 +4183,204 @@ impl Database {
             .optional()?;
         Ok(row)
     }
+
+    /// Apply an incoming sync delta with state-aware conflict resolution
+    /// (#488). Returns `Ok(true)` when the delta was applied or inserted,
+    /// `Ok(false)` when rejected (no-op / regression / forward-incompat).
+    ///
+    /// Conflict rule (in priority order):
+    ///
+    /// 1. Unknown state literal -> log + reject (forward-incompat from a
+    ///    newer peer; do not panic).
+    /// 2. No local row -> insert.
+    /// 3. Tombstone involved (either side has `deleted_at`) -> plain LWW
+    ///    on `updated_at`.
+    /// 4. Local terminal + incoming non-terminal -> REJECT. Terminal is
+    ///    sticky; a peer observing an earlier state must not regress us.
+    /// 5. Both terminal but disagree on state -> keep the row with the
+    ///    later `exited_at`; on tie, deterministic tiebreak on
+    ///    `acquired_by_host` (lexicographic, lower wins).
+    /// 6. Otherwise -> LWW on `updated_at`.
+    pub fn apply_wake_attempt_delta(&self, delta: &crate::sync::WakeAttemptDelta) -> Result<bool> {
+        // Forward-incompat guard: unknown state literal does not panic.
+        let delta_state = match crate::wake_attempts::WakeAttemptState::parse_state(&delta.state) {
+            Ok(s) => s,
+            Err(e) => {
+                eprintln!(
+                    "[legion sync] wake_attempt_delta rejected: unknown state {:?} \
+                         (attempt_id={}, err={})",
+                    delta.state, delta.attempt_id, e
+                );
+                return Ok(false);
+            }
+        };
+
+        let tx = self.conn.unchecked_transaction()?;
+        let local = tx
+            .query_row(
+                "SELECT state, exited_at, updated_at, deleted_at, acquired_by_host \
+                 FROM wake_attempts WHERE attempt_id = ?1",
+                rusqlite::params![&delta.attempt_id],
+                |r| {
+                    Ok((
+                        r.get::<_, String>(0)?,
+                        r.get::<_, Option<String>>(1)?,
+                        r.get::<_, String>(2)?,
+                        r.get::<_, Option<String>>(3)?,
+                        r.get::<_, Option<String>>(4)?,
+                    ))
+                },
+            )
+            .optional()?;
+
+        let signal_ids_json = serde_json::to_string(&delta.signal_ids)?;
+
+        let applied = match local {
+            None => {
+                tx.execute(
+                    "INSERT INTO wake_attempts \
+                     (attempt_id, persona_id, repo_name, signal_ids, state, \
+                      acquired_by_host, acquired_at, spawned_pid, spawned_at, \
+                      exit_observed_at, exited_at, exit_status, outcome, \
+                      deleted_at, updated_at) \
+                     VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11, ?12, ?13, ?14, ?15)",
+                    rusqlite::params![
+                        &delta.attempt_id,
+                        &delta.persona_id,
+                        &delta.repo_name,
+                        &signal_ids_json,
+                        &delta.state,
+                        &delta.acquired_by_host,
+                        &delta.acquired_at,
+                        delta.spawned_pid.map(|v| v as i64),
+                        &delta.spawned_at,
+                        &delta.exit_observed_at,
+                        &delta.exited_at,
+                        &delta.exit_status,
+                        &delta.outcome,
+                        &delta.deleted_at,
+                        &delta.updated_at,
+                    ],
+                )?;
+                true
+            }
+            Some((local_state_str, local_exited, local_updated, local_deleted, local_host)) => {
+                let local_state =
+                    match crate::wake_attempts::WakeAttemptState::parse_state(&local_state_str) {
+                        Ok(s) => s,
+                        Err(_) => {
+                            // Local row is corrupt -- treat any incoming
+                            // well-formed delta as the truth (forward-only
+                            // schema recovery). LWW still applies.
+                            if delta.updated_at <= local_updated {
+                                tx.commit()?;
+                                return Ok(false);
+                            }
+                            self.upsert_wake_attempt_overwrite(&tx, delta, &signal_ids_json)?;
+                            tx.commit()?;
+                            return Ok(true);
+                        }
+                    };
+
+                let tombstone_involved = local_deleted.is_some() || delta.deleted_at.is_some();
+                let local_terminal = local_state.is_terminal();
+                let delta_terminal = delta_state.is_terminal();
+
+                if tombstone_involved {
+                    // Plain LWW on updated_at for tombstones.
+                    if delta.updated_at > local_updated {
+                        self.upsert_wake_attempt_overwrite(&tx, delta, &signal_ids_json)?;
+                        true
+                    } else {
+                        false
+                    }
+                } else if local_terminal && !delta_terminal {
+                    // Terminal-is-sticky: reject the regression. The
+                    // local row keeps its terminal state regardless of
+                    // updated_at -- a newer non-terminal write is a
+                    // happens-before violation by definition.
+                    false
+                } else if local_terminal && delta_terminal {
+                    if local_state == delta_state {
+                        // Both same terminal -- LWW on updated_at to
+                        // accept fresher metadata (outcome label, etc).
+                        if delta.updated_at > local_updated {
+                            self.upsert_wake_attempt_overwrite(&tx, delta, &signal_ids_json)?;
+                            true
+                        } else {
+                            false
+                        }
+                    } else {
+                        // Both terminal but disagree -- keep the later
+                        // exited_at; on tie, deterministic tiebreak on
+                        // acquired_by_host (lower lexicographic wins).
+                        let delta_wins = match (&delta.exited_at, &local_exited) {
+                            (Some(d), Some(l)) if d != l => d > l,
+                            (Some(_), None) => true,
+                            (None, Some(_)) => false,
+                            _ => {
+                                // exited_at tie -- break on host id.
+                                match (&delta.acquired_by_host, &local_host) {
+                                    (Some(d), Some(l)) => d < l,
+                                    (Some(_), None) => true,
+                                    _ => false,
+                                }
+                            }
+                        };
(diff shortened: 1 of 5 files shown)


ssilvius:
**Rust review from smugglr -- LGTM** (self-approve blocked, posting as comment)

**Verified locally** on `origin/feat/487-wake-attempts-schema` (58c52a0):
- `cargo test --bin legion wake_attempt`: 24 passed (FSM module 6 + db 18; PR body says 23, off-by-one is the migration test)
- `cargo clippy --bin legion -- -D warnings`: clean

## Strengths

- Two-layer FSM enforcement is genuine: `can_transition_to` pre-check + SQL `WHERE state = ?from` predicate. Either layer alone is insufficient (Rust catches sync-resolved garbage; SQL catches concurrent moves between read-and-write); having both is the safety property the comment promises.
- Terminal-is-sticky is enforced through the `state IN ('claimed','spawning','running','exiting')` predicate in `record_wake_attempt_outcome`, not just through the transition table. Late-arriving hooks cannot rewrite a settled row, and `record_wake_attempt_outcome_leaves_terminal_rows_alone` pins the exact diagnostic shape (`IllegalWakeAttemptTransition { current: "done" }`, not `WakeAttemptNotFound`).
- The missing-vs-illegal disambiguation is the right call. Conflating them would let a future caller branch on `WakeAttemptNotFound` and retry, silently swallowing corruption. Worth the extra `get_wake_attempt` on the unhappy path.
- `parse_state` instead of `FromStr::from_str` -- the comment is correct that the std `from_str` ambient is a footgun for inherent decoders that produce a domain error.
- `WakeAttemptStateDecodeError` distinct from `WakeAttemptNotFound` carries the same discipline: corruption is not absence.
- Host-scoped `list_local_orphans` with an in-test peer-host fixture confirming the filter actually drops the peer row -- this is the load-bearing safety claim and it's pinned.
- Partial indices keyed on `WHERE deleted_at IS NULL` match smugglr's soft-delete convention; the two indices map to the documented hot paths (crash recovery vs operator history).
- `transition_table_is_exhaustive_check` asserts the count -- accidental table edits fail loudly.
- `try_claim_wake_attempt` mirrors `try_acquire_persona_lease`'s atomic UPDATE-WHERE shape, including `acquired_by_host IS NULL` belt-and-braces alongside `state = 'queued'`.

## Non-blocking observations

1. **`mark_wake_attempt_exit_observed` and `set_wake_attempt_pid` don't gate on state.** A terminal row still has its `updated_at` bumped (and `exit_observed_at` overwritten) by a late stop-hook firing. Per the design intent ("hook is an expediter, not authoritative") this might be fine, but bumping `updated_at` on terminal rows could cause smugglr sync churn once #488 lands. Worth deciding whether to add `AND state IN ('claimed','spawning','running','exiting')` to the WHERE clause or let it ride.

2. **`record_wake_attempt_outcome` skips the Rust-side FSM check.** It's the one path where `can_transition_to` is NOT consulted -- the constraint is enforced only via the SQL `state IN (...)`. The result is identical (the reaper-shortcut transitions are all in the table), but a one-liner pointing this out in the function comment would help a future reader who expects the same two-layer pattern everywhere.

3. **`spawned_pid` round-trips u32 -> i64 -> u32.** On Unix PIDs fit in i32 and the column is `INTEGER`, so this is safe; flagging only because `TryFrom` would surface a programming error rather than silently masking one if a future caller passes a tagged value. Hint-not-authority per the comment, so probably noise.

4. **`signal_ids` as JSON TEXT means no SQL-side `WHERE signal_id = X` queries.** N is small enough that this is the right tradeoff, but if any future consumer needs to ask "which attempt covers signal X?" it'll need a join table or a `LIKE`. Worth a comment if you expect that query shape.

5. **Double `#[allow(dead_code)]`** (module-level on `wake_attempts.rs`, impl-level on the db.rs additions) is the right pragmatic choice given db.rs is mostly live -- noting for the cleanup pass when #488/#489/#490 land: drop both allows in the PR that wires the first consumer.

None of these block. Ready to merge once approval count is satisfied.

ssilvius:
Bundled #488 onto this branch per kessel's PSA (same files: src/db.rs + src/sync.rs, single PR per area beats stacked PRs). New commit 54306cd adds:

- `WakeAttemptDelta` struct in src/sync.rs (after PersonaWakeLeaseDelta).
- `apply_wake_attempt_delta` in src/db.rs with state-aware conflict rule (terminal-is-sticky against non-terminal, both-terminal tiebreak on exited_at then host).
- 8 new tests pinning every conflict-rule branch.

This PR now closes both #487 and #488. 813 tests pass.

Issue #492, opened 2026-05-23
chore: audit LEGION_AUTO_WAKE branches + stop-hook gate decision for watch-pty


ssilvius:
## Goal

**Two related behavior-change risks introduced by the PTY migration:**

1. `LEGION_AUTO_WAKE=1` semantic shift: today it means "this is a non-interactive watch wake, suppress interactive prompts." Under PTY it is a **fully interactive** session that happens to be watch-spawned. Every hook / code path branching on this env var needs an audit and an explicit decision.
2. `plugin/hooks/stop.sh` runs incomplete-task gate + reflect-prompt gate on every CC Stop. Today `-p` sessions hit Stop differently from interactive PTY ones. PTY migration will fire these gates on every wake -- and a blocking gate hitting the 8-block stop-hook cap (CC 2.1.143) will get the agent hard-killed by CC. Reaper observes EOF and continues; not fatal, but operationally noisy.

This issue: audit + decide.

### Audit deliverable

A markdown doc at `docs/decisions/2026-05-watch-pty-env-audit.md` enumerating:

1. Every file under `plugin/hooks/` that reads `LEGION_AUTO_WAKE`. For each: what it does today, whether the behavior is desired under `LEGION_SPAWN_SOURCE=watch-pty`, and the recommended change.
2. Every Rust file under `src/` that reads or sets `LEGION_AUTO_WAKE`. Same per-site analysis.
3. Specific decision for `plugin/hooks/stop.sh`:
   - **Incomplete-task gate** under `watch-pty`: keep / suppress / new behavior?
   - **Reflect-prompt gate** under `watch-pty`: keep / suppress / new behavior?
   - Justification per gate. Default lean: suppress both under `watch-pty` since auto-wake sessions are short-lived and should not be blocked from exiting by the same gates that protect operator-attended sessions. But operator may want the reflect prompt to capture wake-time learnings -- explicit decision required.

### Implementation deliverable

Code changes following the audit decisions:

- Per-site env-var-branch updates (each site either keeps reading `LEGION_AUTO_WAKE` as-is, or switches to `LEGION_SPAWN_SOURCE`, or both)
- `plugin/hooks/stop.sh` updates per the gate decisions
- Tests updated to cover the new behavior

### Behavior

Implementation must match the audit document exactly. Reviewer should be able to read the audit and verify every decision is implemented.

### Error Handling

N/A -- this is a behavior audit + targeted edits, not a new feature surface.

## Out of Scope

[...]

Issue #493, opened 2026-05-23
feat: legion watch session-end CLI + plugin stop-hook handoff


ssilvius:
## Goal

**Optimistic expediter, not authority.** The legion plugin's `stop.sh` calls `legion watch session-end --attempt-id $LEGION_WAKE_ATTEMPT_ID` when the CC session is stopping. This writes `exit_observed_at` on the wake_attempts row, letting the watch reaper skip a poll cycle and converge faster.

**PTY EOF + PID-poll remain the AUTHORITATIVE completion signal.** The hook handoff is a speed-up only -- if the hook never fires (8-block cap, hook write failure, network blip), the reaper still observes EOF and converges. Critical so a misbehaving stop hook cannot wedge a PTY.

### Interface
```rust
// new CLI subcommand wired into main.rs's watch subgroup
legion watch session-end --attempt-id <UUIDv7>
```

```rust
// src/watch.rs
/// Records exit_observed_at on the attempt row. Idempotent. Returns Ok(())
/// even if the attempt is already in a terminal state (the hook fires
/// regardless of state).
pub fn record_session_end(db: &Database, attempt_id: &str) -> Result<()>;
```

### Behavior

- CLI: parses `--attempt-id`, opens the legion DB, calls `record_session_end`. Exits 0 on success, non-zero on error (so the hook log shows the failure but the hook itself does NOT block CC's stop).
- `record_session_end` calls `db.mark_wake_attempt_exit_observed(attempt_id)` (from #487). No-op if the row does not exist (idempotent for re-fires).
- `plugin/hooks/stop.sh` adds, near the end of the script, gated on `$LEGION_WAKE_ATTEMPT_ID` being non-empty:
  ```bash
  if [ -n "${LEGION_WAKE_ATTEMPT_ID:-}" ]; then
    legion watch session-end --attempt-id "$LEGION_WAKE_ATTEMPT_ID" >/dev/null 2>&1 || true
  fi
  ```
  The `|| true` is intentional -- hook MUST NOT block stop on this expediter.
- Watch reaper: in `reap_finished`, BEFORE the PTY/Child exit poll, optionally read `db.get_wake_attempt(attempt_id).exit_observed_at`. If set AND the PTY has not exited within a short grace window (e.g., 30s), kill the PTY. This is the "hook-says-done but PTY hung" path -- still authority-of-EOF, just with a kill nudge after a grace period.

### Error Handling

[...]

PR #500, opened 2026-05-23
chore(#492): audit phase -- LEGION_AUTO_WAKE + stop.sh gate decisions

ssilvius:
First commit of #492. Audit-only deliverable; the implementation phase waits on Sean signing off on the gate decisions + #489 landing.

## Audit findings

- `LEGION_AUTO_WAKE` is **set** in `src/watch.rs:796` and **read** by zero call sites inside legion code today. Keep it for external backwards compatibility; the audit-relevant branch point becomes a new `LEGION_SPAWN_SOURCE=watch-pty` marker that #489 will set.
- Zero hooks under `plugin/hooks/` read `LEGION_AUTO_WAKE` -- the env var is functionally a no-op marker today.
- `plugin/hooks/stop.sh`'s two gates (incomplete-task block, reflect-prompt) are calibrated for operator-attended sessions. Under watch-pty both should suppress because (a) the 8-block stop-hook cap in CC 2.1.143 hard-kills the agent if the gate fires repeatedly, (b) watch wakes are atomic units that need to exit cleanly so the orchestration layer can decide next steps.

## Proposed mechanism

Add a single early-exit at the top of `stop.sh` (right after the existing `LEGION_SKIP_STOP_BLOCK` bypass) that exits 0 when `LEGION_SPAWN_SOURCE=watch-pty`, logging the suppression to `bypass.jsonl` for operator audit. Full snippet in the audit doc.

## What's in this commit

- `docs/decisions/2026-05-watch-pty-env-audit.md` -- the audit doc + decisions + implementation checklist.

## What's NOT in this commit yet (gated on Sean sign-off + #489 landing)

- The `stop.sh` edit.
- The `src/watch.rs` PTY-branch env-var stamp (lands as part of #489).
- Integration tests for both gate paths.
- `docs/site/getting-started.md` env-var table update.

Part of #492. Part of #495.

### DIFF
--- a/docs/decisions/2026-05-watch-pty-env-audit.md
+++ b/docs/decisions/2026-05-watch-pty-env-audit.md
@@ -114,5 +114,18 @@ The `LEGION_SKIP_STOP_BLOCK=1` bypass is kept as an orthogonal escape for diagno
 ## Out of scope for #492
 
 - The PTY spawn branch itself (#489).
-- The `legion watch session-end` CLI (#493) -- lands after this audit's gate decisions so the wire-up uses the env var settled on here.
 - Forward-compatibility for `LEGION_AUTO_WAKE=0` -- the env var has never been read by legion code; no callers to migrate.
+
+## Bundled work: #493 stop.sh handoff
+
+Per kessel's bundle doctrine, the `stop.sh` half of #493 (the session-end CLI handoff) lands on the same branch as this audit -- both edit `plugin/hooks/stop.sh`. The CLI subcommand itself (`legion watch session-end --attempt-id <id>`) requires the `mark_wake_attempt_exit_observed` method from #487 and lands in the watch.rs bundle PR alongside #489/#490/#491.
+
+The handoff is forward-compatible: the call is gated on `$LEGION_WAKE_ATTEMPT_ID` being set, the binary call is `|| true`-suppressed, and the CLI subcommand does not exist yet -- the hook is dead code until the consumer lands. The shell-test stub records every CLI invocation so we can assert call shape without a running binary.
+
+Hook firing rules (which paths invoke `session_end_handoff`):
+- `LEGION_SKIP_STOP_BLOCK=1` bypass: yes (explicit operator session-end).
+- `LEGION_SPAWN_SOURCE=watch-pty` bypass: yes (every watch-pty wake records its exit).
+- Reflect-prompt fired previously (marker exists): yes (clean exit path).
+- No work marker (session had no real work): yes (clean exit path).
+- Incomplete-task gate blocks: no (agent staying alive; would mis-mark exit).
+- Reflect-prompt block fires: no (agent staying alive).
--- a/docs/decisions/2026-05-watch-pty-env-audit.md
+++ b/docs/decisions/2026-05-watch-pty-env-audit.md
@@ -1,0 +1,118 @@
+# Watch PTY env-var + stop-hook audit
+
+**Date:** 2026-05-23
+**Issue:** #492 (audit phase of the #495 PTY migration arc)
+**Status:** Audit phase complete. Implementation phase pending operator sign-off + #489 landing.
+
+## Why this audit exists
+
+The PTY migration (#495) changes two background assumptions:
+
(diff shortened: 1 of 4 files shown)


ssilvius (commented):
## Verdict: LGTM

Forward-compatible audit-and-implement PR. The env-var gate is wired but dead until #489 stamps `LEGION_SPAWN_SOURCE=watch-pty`, which is the right serial-shipping shape. Audit doc is thorough -- enumerates the one set-site (`src/watch.rs:796`) and zero read-sites for `LEGION_AUTO_WAKE`, then justifies suppression of both stop.sh gates under watch-pty with the 8-block cap + cooperative-wake rationale.

Verified locally on `origin/feat/492-watch-pty-env-audit`:
- `bash plugin/hooks/test-stop-task-block.sh` -> 12/12 passed (was 10/10 on main, +2 for the new watch-pty branch).
- The negative test (`LEGION_SPAWN_SOURCE=manual-test` still blocks) pins the exact-match contract -- any future change that relaxes the comparison will fail this test.
- Ordering after the `LEGION_SKIP_STOP_BLOCK` bypass is correct: the two escapes are orthogonal (one is per-session diagnostic, the other is a spawn-source routing decision), and the audit doc explicitly carves that out.
- Distinct telemetry `pattern: "watch-pty-skip"` keeps the bypass log greppable from operator-issued `LEGION_SKIP_STOP_BLOCK` bypasses.

## Non-blocking observations

1. **PR body is stale against commit 2.** The "What's NOT in this commit yet" section lists "The stop.sh edit", "Integration tests for both gate paths", and "docs/site/getting-started.md env-var table update" -- all three landed in `26ee312`. Only the `src/watch.rs` PTY-branch env stamp is correctly excluded (still gated on #489). Worth a body update so reviewers reading just the description see the actual scope.

2. **Drive-by reflection-text test pin is outside #492 scope.** The `assert_contains` updates from `'What would you tell another agent'` -> `'Drop one thing a teammate would not have known walking in cold'` fix tests that were stale against the current `stop.sh` reflection prompt. Good fix, but it's not a #492 deliverable -- worth a sentence in the commit body so future blamers don't think it's part of the watch-pty change.

3. **Env-var doc row doesn't pin the recognized value.** `docs/site/getting-started.md` says "Set to `watch-pty` by the watch daemon on PTY-spawned wakes" -- the implication that only the exact string `watch-pty` is recognized is implicit. Adding a sentence like "Only the literal value `watch-pty` triggers the bypass; any other value is treated as unset" prevents a future contributor from extending the comparison without re-reading the test.

4. **`LEGION_AUTO_WAKE` retained as a no-op marker.** Audit confirms zero in-tree readers. Worth filing a follow-up to deprecate + remove once watch-pty is the default and a deprecation window has passed -- the audit doc's "Out of scope" section already gestures at this but doesn't file the ticket.

5. **Telemetry call is fire-and-forget** (`2>/dev/null || true`) matching the existing `LEGION_SKIP_STOP_BLOCK` pattern -- correct (the hook must not block on telemetry write failure), but it means bypass-log loss is silent. Acceptable trade.

Nothing blocking. Ship it.

ssilvius:
Bundled #493 stop.sh handoff onto this branch per kessel's PSA (same file: plugin/hooks/stop.sh). Commit 3931564 adds:

- `session_end_handoff()` function at top of stop.sh (defined before any caller).
- Invocations on every clean-exit path: LEGION_SKIP_STOP_BLOCK bypass, watch-pty bypass, reflect-marker-exists, no-work-marker. NOT on block paths (agent staying alive).
- 3 new shell tests: stub now logs invocations to $LEGION_STUB_LOG; asserts handoff fires under watch-pty + LEGION_WAKE_ATTEMPT_ID, and is gated on LEGION_WAKE_ATTEMPT_ID. 15/15 pass.
- Audit doc updated with bundling rationale + firing-rule table.

The CLI subcommand half of #493 (`legion watch session-end`) requires `mark_wake_attempt_exit_observed` from #487 -- lands in the watch.rs bundle PR alongside #489/#490/#491. Until then the `|| true` makes this dead code.

PR now closes #492 + half of #493.

Issue #536, opened 2026-05-29
fix: cluster sync never runs under the daemon and transfers no data


ssilvius:
## Goal

**Make cluster sync actually run and transfer data between nodes.** Today `legion cluster init/enable` configures sync correctly, but no data ever syncs between peers -- and under normal operation the sync actor never even starts.

This was found while bringing up a two-node LAN cluster: both nodes configured with the same key, `cluster status` reporting `enabled: yes`, yet `last_sync: never` and zero peers discovered. Tracing the code surfaced three independent defects, listed below in priority order.

### Defect 1 (blocker) -- the daemon never starts the sync actor

The plugin runs legion via `legion daemon`, whose watch loop is `daemon.rs::run_watch_task` (src/daemon.rs:283). That loop does health/poll/wake but never loads `cluster.toml` or spawns the sync actor. The actor is only spawned by the *foreground* `legion watch` loop:

```rust
// src/watch.rs:1674 -- present in `legion watch`, ABSENT in daemon.rs::run_watch_task
let _sync_handle = match crate::cluster::ClusterConfig::load(data_dir) {
    Ok(cc) if cc.enabled && cc.secret.is_some() => {
        eprintln!("[legion watch] cluster sync enabled on port {} ...", cc.port, ...);
        match crate::sync_actor::SyncHandle::spawn(data_dir, cc) { ... }
    }
    _ => None,
};
```

Compounding it: `legion watch` prints `is deprecated -- use legion daemon` on startup. So the only code path that starts cluster sync is the deprecated one; the supported path (daemon) silently skips it. Confirmed empirically: under the daemon, nothing binds UDP 31337 (`ss -lunp` shows no listener). Running `legion watch` in the foreground binds the socket immediately.

**Required behavior:** `daemon.rs::run_watch_task` must perform the same `ClusterConfig::load` + `SyncHandle::spawn` as `watch.rs`, and hold the resulting handle for the lifetime of the watch task. When the daemon runs with `cluster.toml` enabled + secret set, the daemon process must bind UDP `<port>` and log `cluster sync enabled on port <port>`.

### Defect 2 (blocker) -- delta transmission is unimplemented

Even via the foreground path, the sync loop discovers peers and computes deltas but never sends them. src/sync_actor.rs:165:

```rust
// TODO: Actually broadcast the deltas to peers
// For now, just log that we would send them
// The wire protocol uses DeltaPacket with upserts/deletes
last_sync = chrono::Utc::now().to_rfc3339();
```

[...]

PR #624, opened 2026-06-11
fix(#536): cluster sync runs under the daemon and actually transfers data

### LINKED ISSUE #250: feat: Tantivy reindex after sync
## Summary

Rebuild Tantivy index after receiving new reflections via sync.

## Parent epic

#244 (multi-node sync)

## Dependencies

- #249 (sync actor)
- #106 (incremental index sync - already accepted)

### A. Incremental add (preferred)
When sync delivers new reflections, add each to index immediately:
```rust
for reflection in incoming_reflections {
    db.insert_reflection(&reflection)?;
    search_index.add_document(&reflection)?;
}
search_index.commit()?;
```

### B. Background reindex
After sync batch completes, trigger async reindex:
```rust
tokio::spawn(async move {
    search::reindex_all(&db, &index)?;
});
```

## Acceptance criteria

- [ ] New reflections from sync are searchable via `legion recall`
- [ ] No manual `legion reindex` needed after sync
- [ ] Index remains consistent if sync is interrupted

ssilvius:
## Summary

Cluster sync now actually syncs (#536). Three defects, one cause-chain: the daemon never spawned the sync actor (only the deprecated foreground watch did), the actor computed deltas but transmission was a TODO, and last_sync lived in memory so cluster status always read "never". The fix switches transmission onto smugglr-core 0.4.0 primitives (DeltaPacket/split_delta, XChaCha20 maybe_encrypt/maybe_decrypt, ReplayGuard), adds the missing receive-side LWW apply methods for reflections/cards/schedules, unifies the spawn behind one shared gate both drivers call, and persists last_sync to cluster.toml only after fully-successful cycles.

### 1. All tests pass

979 unit + 161 integration tests green locally (cargo test), clippy --all-targets -D warnings clean, fmt clean. New coverage: 6 apply-LWW tests in src/db.rs (insert-when-missing, newer-overwrites/older-noop, tombstone-wins for reflections; insert-then-LWW for cards and schedules; effective_sync_ts picks latest of created/updated/deleted), and 6 actor tests in src/sync_actor.rs (spawn gate returns None when disabled or secretless, packet seal/unseal round trip through real encryption, apply_packet round-trips a reflection into the DB, unknown table applies zero, LWW does not clobber newer local rows, persist_last_sync survives ClusterConfig reload).
Evidence: src/db.rs tests apply_reflection_delta_inserts_when_missing through effective_sync_ts_picks_latest_of_three; src/sync_actor.rs tests spawn_gate_none_when_disabled_or_secretless through apply_packet_lww_does_not_clobber_newer_local

### 2. Simplify pass completed

Recorded clean on HEAD. Substantive call made during the pass: the three apply_*_delta bodies stay parallel rather than unified -- the 2026-06 audit adversarially verdicted the equivalent delta-getter shape as coincidence-not-duplicate, and unifying them requires a column-driven query builder (explicit operator constraint). What IS one invariant got extracted: effective_sync_ts (the LWW comparison, one fn + test), pack_table (generic over any serde delta), spawn_sync_if_enabled (one gate, two callers). The three hand-rolled *_to_map converter fns and their test were deleted outright -- the delta structs already derive Serialize, so serde_json::to_value replaces 148 lines.
Evidence: quality gate row on HEAD f65d03d; diff shows reflection_to_map/card_to_map/schedule_to_map deleted

### 3. Review pass completed and issues fixed

Self-review against the issue's error-handling requirements: config errors and spawn failures log and return None so the watch loop always runs (spawn_sync_if_enabled error arms); send failures mark the cycle unsuccessful so last_sync does not advance and the window retries (all_sent flag); a failed peer logs and the loop continues; receive-side failures (bad decrypt, bad packet, db open, per-row apply) each log and continue -- nothing crashes the actor. The expect() on Runtime::new in the old code was also replaced with a log-and-return (no-panic discipline). Team review requested via signals on this PR.
Evidence: src/sync_actor.rs spawn_sync_if_enabled error arms, all_sent gating in run_sync_loop, receive_loop error arms

### 4. PR created and linked to this issue

This PR, branch fix/536-cluster-sync, linked to issue #536 and kanban card 019e7617-33b9-7f92-a0d6-6995eb21edb1.
Evidence: PR metadata; pr create --task linkage

## Not done

- Gossip-stack adoption: smugglr-core's multicast::Gossip is coupled to smugglr's own LocalDb/Config; modeling legion tables as a smugglr datasource is a redesign, not this bug fix. The broadcast primitives were the right cut.
- Syncing the remaining delta types (rate-limit/usage samples, uncertainty, wake_attempts) over this transport: the issue scopes transmission to reflections/cards/schedules/leases; the other deltas keep their existing paths and can join the packet loop in a follow-up by adding pack_table lines.
- Tantivy/embedding refresh for peer-applied reflections: applied rows have no embedding (by design, each node computes its own) and are not in the search index until reindex/backfill runs; surfaced previously as closed #250, re-file if live clusters show recall gaps.
- Cross-subnet discovery (LAN broadcast only, per issue out-of-scope) and un-deprecating foreground watch.
- A live two-node integration test: needs two daemons and UDP broadcast in CI; the loopback-free unit seam (pack -> seal -> unseal -> apply) covers the protocol path deterministically instead.

### DIFF
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -4,6 +4,14 @@
 //! and synchronize reflections, cards, and schedules using delta packets.
 //! All traffic is encrypted with XChaCha20-Poly1305 using a pre-shared key.
 //!
+//! Trust model: the PSK is the entire boundary. Schedules sync across the
+//! cluster and the schedule runner executes their commands, so possession
+//! of the secret (or compromise of any one node) is command execution on
+//! every node. Share the key accordingly. Discovery announcements are
+//! unauthenticated by design (availability only -- delta payloads stay
+//! AEAD-sealed); an on-LAN spoofer can misdirect traffic but not read or
+//! forge it.
+//!
 //! Configuration is stored in `<data_dir>/cluster.toml`:
 //! ```toml
 //! enabled = true
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -486,6 +486,12 @@ async fn run_watch_task(data_dir: &Path) {
 
     let host = watch::resolve_host_id();
 
+    // Spawn the cluster sync actor when enabled (#536 defect 1: this task
+    // previously never loaded cluster.toml, so sync only ran under the
+    // deprecated foreground `legion watch`). The handle lives for the
+    // lifetime of this task; sync is optional and never fatal.
+    let _sync_handle = crate::sync_actor::spawn_sync_if_enabled(data_dir, "[legion daemon]");
+
     // Timer intervals are read before `config` moves into the WatchLoop.
     let poll_interval = std::time::Duration::from_secs(config.poll_interval_secs);
     let health_interval = std::time::Duration::from_secs(config.health_poll_secs);
--- a/src/db.rs
+++ b/src/db.rs
@@ -4025,6 +4025,262 @@ impl Database {
         tx.commit()?;
         Ok(late_loser)
     }
+
+    /// Apply a peer's reflection delta with last-write-wins merge (#536).
+    ///
+    /// No local row: INSERT (embedding stays NULL -- each node computes its
+    /// own embeddings, see the ReflectionDelta doc). Local row exists: the
+    /// delta wins only when its effective timestamp (max of updated_at /
+    /// deleted_at, falling back to created_at) is strictly newer than the
+    /// local row's. Tombstones ride the same comparison.
+    pub fn apply_reflection_delta(&self, delta: &crate::sync::ReflectionDelta) -> Result<()> {
+        let tx = self.conn.unchecked_transaction()?;
+
+        let local: Option<(String, Option<String>, Option<String>)> = tx
+            .query_row(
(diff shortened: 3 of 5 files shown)


ssilvius:
Review pass completed post-open (legion-review pipeline, 2 dimension agents, 17 findings). Both dimensions independently caught a HIGH with crate-source proof: split_delta parts share one seq, ReplayGuard drops repeats -- any batch over ~1.4KB silently lost rows after part 0 while last_sync advanced. Fixed in 7fde3c4: per-part seq restamp + regression test (multipart_batches_pass_a_fresh_replay_guard), epoch-millis seq seed (restart outage), refuse-to-run-without-key, receive-loop single conn + busy_timeout, recv backoff, fail-loud sends, checked port arithmetic, trust-model docs. Three MEDs dispositioned with rationale (LWW tiebreaker + ack/anti-entropy = scope-excluded policy changes; daemon spawn placement matches the issue's literal Required-behavior). Full disposition in the legion-review quality-gate details on HEAD 7fde3c4.

Issue #609, opened 2026-06-11
refactor(B): db stream -- db.rs into db/ module tree (kessel recipe, one extraction PR)


ssilvius:
## Why

db.rs is 8,815 lines: ~120 methods across 7 impl blocks, a 700-line init_schema owning ~18 tables in patch order, a 3,700-line test module. The split is low-risk: impl-Database-spans-files is already proven in-repo (documents.rs, uncertainty/storage.rs) and conn is already pub(crate). Evidence: audit document 019eb3b9-6674-79b1-b0a9-82eab17f9a21.

## Scope (serial; #606 lands first)

0. Predecessor: #606 (get_reflections_by_ids column shift + REFLECTION_COLUMNS const).
1. ONE extraction PR per the kessel recipe (reflection 019e7c5a): db/mod.rs = infra only (Database, open, has_column, thin init_schema dispatcher calling per-domain create_tables in one transaction, migrate_* after); domain files own DDL + methods + tests (reflections, board, kanban, schedules, sync, wake/leases, health, statusline_samples, stats, quality_gates, autonomy, audit, scip, heartbeat); db/testutil.rs collapses the four temp-db helpers.
2. In-flight DRY limited to verified true duplicates: active-team-post WHERE fragment (6 sites -> const in db/board.rs), REFLECTION_COLUMNS usage at all 11+ sites. Tasks+cards land side by side in db/kanban.rs UNCHANGED (collapse is a follow-on).

## Acceptance criteria

- [ ] db.rs deleted; db/ tree compiles with zero public-API change (no caller edits outside use paths)
- [ ] init_schema dispatcher: per-domain DDL, one transaction, migrations after
- [ ] Four test-db helpers converged on db/testutil.rs
- [ ] cargo test green before and after; clippy --all-targets -D warnings
- [ ] Parent epic: #607

ssilvius:
Stream B complete in PR #629: db.rs (9,329 lines) -> db/ tree, mod.rs infra-only + 18 domain files + testutil, one provably-mechanical extraction (line-coverage-asserted carve), init_schema one-transaction semantics preserved, ACTIVE_TEAM_POST_WHERE const, test count 1152->1153. Review pipeline caught and fixed a TempDir-drop bug in the converged test helper pre-merge.

Issue #607, opened 2026-06-10
epic: audit-driven refactor 2026-06 -- four god files into module trees, verified DRY only


ssilvius:
## Why

53.8k lines of Rust; db.rs (8,815) + main.rs (8,130) + integration.rs (6,696) + watch.rs (4,767) hold 53% of it. The 2026-06 audit (8 sym-first explorers, adversarial verification of every must-change-together claim) produced 94 findings: 24 high, 19 verified true duplicates, 10 overturned as coincidence. Full report: docs/plans/2026-06-refactor-audit.md. Per-subsystem evidence: legion document list --doc-type audit-finding (8 documents, 2026-06-10).

## Stream DAG (kessel rules: serial within a stream, streams interleave, mechanical moves never mix with semantic changes)

- A tests stream -- gates B, C, D (the net before the trapeze)
- B db stream -- #606 bug fix first, then ONE extraction PR
- C cli stream -- in-place DRY -> datadir -> handler files, serial PRs
- D watch stream -- #536 fix shape decided first, then split
- E comms stream -- compose/validate entry point (closes MCP #587 bypass), then mcp/ split
- F serve stream -- ServeError -> mount channel::router -> schedule background task
- G hooks stream -- shell lib/ + generation merge + dead deletion (independent, anytime)
- H midtier cleanup -- plugin_call helper + list-policy fix + dead code (trails everything)

## Deliberately NOT filed

- Crate split: answered with compile-time data after the module trees land. Default remains modules-not-crates.
- tasks/cards CRUD collapse: follow-on filed only after db/kanban.rs makes the shared-table invariant visible.
- All 33 low-severity findings: recorded in the audit documents, recoverable, not work items.

## Acceptance criteria

- [ ] docs/plans/2026-06-refactor-audit.md committed
- [ ] Streams A-H complete or explicitly re-scoped with rationale
- [ ] No PR in this epic mixes a mechanical move with a semantic change

ssilvius:
Operator decision 2026-06-10: no crates, final. The epic body's 'answered with compile-time data after the trees land' framing is superseded -- the module trees ARE the end state. Definition-layer concept/shape preserved as substrate documents before the board reset: charter 019eb40e-8cd5, spec-format decisions 019eb40e-8cde, service-design formats 019eb40e-8ce6 (legion document list --doc-type concept --surface definition-layer).

ssilvius:
[...]

Issue #606, opened 2026-06-10
bug: get_reflections_by_ids SELECT omits updated_at -- column shift panics serve search


ssilvius:
## Found by

2026-06 refactor audit (db explorer, adversarially verified as part of cluster reflection-column-list).

## Summary

map_reflection_row (src/db.rs:85-98) positionally reads 11 columns: updated_at at index 4, parent_id at index 10. get_reflections_by_ids (src/db.rs:1866-1867) SELECTs only 10 columns, omitting updated_at. Every field from index 4 on shifts by one: audience TEXT loads into updated_at, and get(10) indexes past the row -- rusqlite InvalidColumnIndex at runtime.

## Blast radius

Sole caller is the serve search path (src/serve.rs:1233), so the dashboard search endpoint is the trigger.

## Fix

- Correct the SELECT to the canonical 11-column list.
- Introduce a REFLECTION_COLUMNS const (pattern precedent: CARD_COLUMNS at src/db.rs:2702) and use it at the 11+ sites that repeat the column list verbatim, so the lists cannot drift again.
- Add a test through get_reflections_by_ids (none exists today -- which is why this survived).

## Acceptance criteria

- [ ] get_reflections_by_ids returns correctly populated Reflection structs (updated_at and parent_id land in their fields)
- [ ] REFLECTION_COLUMNS const exists and is used by every reflections SELECT paired with map_reflection_row
- [ ] Regression test covers get_reflections_by_ids round-trip
- [ ] Serve search endpoint exercises the path without error

ssilvius:
Shipped in PR #623: REFLECTION_COLUMNS const at all 16 map_reflection_row sites, column-shift panic fixed, regression tests in src/db.rs.

Issue #376, opened 2026-04-29
bug: bullpen posts have no decay -- filter stale at injection, not at policy


ssilvius:
## Symptom

When a long-dormant agent wakes, it receives its full unread bullpen queue as channel notifications and processes them oldest-first, posting replies to questions that are days or weeks old, already answered, no longer relevant, or where the asker has long since moved on.

Observed 2026-04-29: platform agent woke and posted ~8 sequential replies to bullpen posts with UUIDv7 prefixes ranging from `019cd22e` (days old) to `019d0291` (recent).

## Root cause

Stale bullpen posts get injected into the agent's conversation context via `notifications/claude/channel`. The agent then pays tokens reading the stale content and deciding what to do with it. Even if it decides "do not reply," it has already cost tokens to reason about a thread that has no live value.

The fix is **filter at injection, not at policy**. The agent must never see a stale post in the first place. Client-side `--reply-stale` flags or "are you sure?" gates still leak the read cost.

## Design

Server-side decay on the bullpen surface:

1. **Default TTL by tag/domain at insert time:**
   - `domain:identity`, `tags:doctrine` -> evergreen (no decay)
   - `tags:design,architecture` -> 14 days
   - default feed post -> 48 hours
   - `is_signal=true` (directed signal) -> 7 days

2. **Strict server-side filter on read.** `legion bullpen --repo X` and the channel-notification feed both apply `WHERE expires_at > now() OR evergreen=true` before any output reaches the agent. Stale posts are not in the result set. The agent has no awareness they exist unless it explicitly asks.

3. **Explicit catchup is opt-in.** `legion bullpen --include-stale` for human review (Sean checking what got missed). Agents have no auto-trigger for this surface.

4. **No client-side reply gates.** Drop the `--reply-stale` flag entirely. If the post is past TTL it is invisible; nothing to reply to.

## Schema

- Add `expires_at` (timestamp, computed at insert from tag-defaults if not provided)
- Add `evergreen` (bool, opt-in for doctrine/identity)
- Index `(repo, expires_at)` for fast filter

## Migration

Backfill `expires_at = created_at + 48h` for existing posts; mark identity/doctrine evergreen by tag scan. Anything past TTL on migration day becomes invisible to the agent surface immediately. Stays in the DB; just not surfaced.

## Acceptance

- `expires_at` populated on every new post
- `legion bullpen` defaults to non-stale; `--include-stale` for explicit review
- Channel notifications never include past-TTL posts
[...]

PR #629, opened 2026-06-11
refactor(#609): split db.rs into a db/ domain module tree

ssilvius:
## Summary

Splits the 9,329-line `src/db.rs` into a `src/db/` domain module tree per the kessel recipe (reflection 019e7c5a, audit 019eb3b9). `db/mod.rs` keeps infrastructure only -- the `Database` struct (`conn` stays `pub(crate)`), `open`, `has_column`, a thin `init_schema` dispatcher, and `rename_repo` (cross-table admin). Sixteen domain files each carry their own `impl Database` block, their DDL, and their tests: reflections, board, sessions, kanban, schedules, sync, wake, health, statusline_samples, stats, quality_gates, autonomy, audit, scip, heartbeat, plus DDL-only documents and uncertainty files whose methods already lived in `src/documents.rs` / `src/uncertainty/storage.rs`.

The extraction was performed by a line-range carve script with a coverage assertion: every line of the old db.rs was either copied verbatim into exactly one new file, blank, or on an explicit discard list (the rewritten scaffolding: import header, impl-block braces, init_schema signature, the test-mod shell). The mechanical-move commit (9d30b5f) therefore contains no logic edits beyond the three declared ones below. Three commits: the move, the test-helper convergence, and an upgrade-path test.

Declared in-flight changes (the two sanctioned by the issue, plus the dispatcher restructure the issue specifies):
1. `ACTIVE_TEAM_POST_WHERE` const in `db/board.rs` replaces the hand-copied active-team-post WHERE fragment at five sites (`get_board_posts_filtered_full`, `get_board_posts_since`, `get_and_mark_unread_board_posts`, `get_unread_count`, `get_recent_board_posts`). The sixth site, `get_unhandled_signals_for_repo`, keeps its `r.`-aliased copy inline because it aliases the table and deliberately has no archive clause; the const's doc comment points at it.
2. `REFLECTION_COLUMNS` (landed in #606) verified, not regressed: all 18 SELECT sites paired with `map_reflection_row` use the const (14 in reflections.rs/board.rs queries plus get_high_value_cross_repo in stats.rs -- see the grep in the evidence below).
3. `init_schema` restructured into per-domain `create_tables` (one transaction) + per-domain `migrate` (after, outside it), preserving each domain's internal patch order.

`Self::has_column` became `Database::has_column` inside the extracted free `migrate` fns (count-asserted substitution); `get_unread_count` switched from a plain `prepare` string to `format!` so it could take the const (same SQL text, covered by five existing tests).

### db.rs deleted; db/ tree compiles with zero public-API change (no caller edits outside use paths)

`src/db.rs` is deleted; `src/db/` holds mod.rs plus 18 domain files. No caller file changed at all -- not even use paths: `db/mod.rs` re-exports the names callers import (`AuditInput`, `CardTimestamp`, `Reflection`, `ReflectionMeta`, `validate_hhmm`), and `impl Database` blocks spanning files keep every method reachable as before. The only files outside `src/db/` touched in this PR are the three test modules in the helper-convergence commit. Re-exports that no caller referenced (the compiler flagged them unused: `AuditEntry`, `QualityGateRow`, `WatchHeartbeat`, `PersonaWakeLease`, `Schedule`, `compute_next_run`, `ReflectionWithEmbedding`, `DashboardRepoStats`, `RepoStats`, `TombstoneCleanupResult`) were dropped rather than kept dead; those types remain `pub` in their domain files and reach callers through method signatures as they always did.
Evidence: `git diff main --stat -- ':!src/db' ':!src/db.rs'` shows only src/documents.rs, src/stats.rs, src/uncertainty/storage.rs (test mods); clean `cargo check --all-targets` with zero non-db edits.

### init_schema dispatcher: per-domain DDL, one transaction, migrations after

`db/mod.rs::init_schema` opens one `unchecked_transaction`, calls the 15 per-domain `create_tables` functions inside it, commits, then runs the four `migrate` functions (reflections, board, kanban, schedules) after and outside it. Each domain's migrate keeps its original patch order; cross-domain order is safe because every migration touches only its own tables and the bullpen-lifecycle backfill (#376) runs in `board::migrate` after `reflections::migrate` has added the columns it reads. Migration 15's single five-index batch was split into its three owning domains.
Evidence: src/db/mod.rs:102-126 (`init_schema`); test `db::tests::init_schema_migrates_a_v1_database` builds a v1-shape database and reopens it through the dispatcher, asserting the migration-14 backfill ran and current read/write paths work; test `db::tests::partial_indexes_created_for_soft_delete` proves all five split partial indexes still exist.

### Four test-db helpers converged on db/testutil.rs

`db/testutil.rs` holds the single `#[cfg(test)] pub(crate) fn test_db() -> Database` (the shape db.rs's 185 tests always used). The other three copies are gone: `documents.rs::open_test_db` (leaked its tempdir), `stats.rs::test_db` (`(Database, TempDir)`), `uncertainty/storage.rs::test_db` (`(TempDir, Database)`); all 20 of their call sites now import `crate::db::testutil::test_db`. The dropped-TempDir shape is safe because `Database::open` writes the schema (creating the WAL sidecars) while the directory still exists, and the held file descriptors outlive the unlink -- the behavior the db tests have exercised since the helper was written.
Evidence: src/db/testutil.rs; commit 16f6c77 diff (3 files, -47/+24); the documents/stats/uncertainty test suites pass in the final `cargo test` run.

### cargo test green before and after; clippy --all-targets -D warnings

Baseline on the branch point: 1152 `#[test]` functions (185 in db.rs), suite green. After: 1153 (186 in src/db/ -- the one addition is `init_schema_migrates_a_v1_database`), suite green at every one of the three commits: 984 + 166 passed, 0 failed, 2 ignored (pre-existing). `cargo clippy --all-targets -- -D warnings` and `cargo fmt -- --check` clean. The `#[allow(dead_code)]` markers that the old block-level attributes provided (`apply_wake_attempt_delta`, `upsert_wake_attempt_overwrite`, the wake-attempt impl block) were re-applied at their new locations with the original "wired by #488/#489/#490" comments.
Evidence: test runs above; quality gate 019eb77b-7689-7f80-b135-ea4d494dbc8b records the simplify pass on HEAD.

### Parent epic: #607

This is stream B of the refactor epic. The kanban task/card dual-CRUD lands side by side in `db/kanban.rs` unchanged, exactly as the issue scopes it -- the collapse is a follow-on under the epic.
Evidence: src/db/kanban.rs carries both `// --- Task CRUD ---` and `// --- Card CRUD (kanban) ---` sections verbatim.

## Not done

- No param renumbering to fully unify the decay/resolved SQL clause (`(evergreen = 1 OR expires_at > ?N) AND resolved_at IS NULL`): the five sites bind different parameter indexes, and renumbering is a semantic change outside a mechanical PR. The const covers the byte-identical base fragment.
- `get_unhandled_signals_for_repo` keeps its `r.`-aliased filter copy (no archive clause, table alias); documented on the const.
- Tasks/cards dual-CRUD collapse, the positional-param builder, and stringly-typed card status fixes from the audit: all follow-ons under epic #607, untouched here.
- The leases + wake_attempts file came in at 1,277 lines as one `db/wake.rs` (under the ~1500 guidance), so it was not split into two files.


ssilvius:
Review outcome: 1 upheld finding, 1 fixed, 0 dispositioned, 0 refuted.

**Fixed (MED, src/db/testutil.rs:10, commit 3ccc933):** the converged test_db() dropped its TempDir before returning, unlinking the db/-wal/-shm files under the live connection -- it worked only through unlinked-open-fd semantics, and the convergence commit had erased the three deleted helpers' comments that guarded against exactly this. The helper now leaks the TempDir via std::mem::forget (the old documents.rs approach) and restores the warning as a doc comment: tests that reopen the path, open a second connection, or checkpoint must bind their own tempdir (as the v1 migration test in db/mod.rs does). This supersedes the PR body's original "dropped-TempDir shape is safe" rationale for the testutil criterion.

Post-fix verification: cargo test green (984 + 166 passed, 0 failed, 2 ignored pre-existing), cargo clippy --all-targets -- -D warnings clean, cargo fmt --check clean. Simplify and pr-write gates re-recorded on the new HEAD. Awaiting merge approval -- not merging.

PR #686, opened 2026-06-23
fix(#685): cfg(unix)-gate the ETXTBSY retry, fix vacuous macOS test

### LINKED ISSUE #685: Review follow-up to #683: cfg(unix)-gate the ETXTBSY retry, fix vacuous macOS test
## Summary

Follow-up to the #683 (#682) code review, which approved with two MED findings that
landed on main and should be fixed forward.

## Findings to fix

1. **Windows misidentification (MED).** `const ETXTBSY: i32 = 26` and
   `spawn_with_etxtbsy_retry` are ungated, but `windows-latest` compiles and runs the
   crate. On Windows `raw_os_error()` returns Win32 codes, where 26 is
   `ERROR_SHARING_VIOLATION` -- a Windows spawn failing with 26 would be misread as
   ETXTBSY and retried for the full 2s budget. ETXTBSY is a POSIX errno; the retry must
   be unix-only.

2. **Vacuous macOS test (MED).** `retries_past_etxtbsy_until_writer_closes` passes for the
   wrong reason on macOS: Darwin raises ETXTBSY only for native Mach-O binaries, not
   shebang-interpreted scripts, so the held write fd never triggers the retry and the
   first spawn succeeds immediately. The test is load-bearing only on Linux.

## Acceptance criteria

- [ ] `ETXTBSY` const + retry consts + `spawn_with_etxtbsy_retry` are `#[cfg(unix)]`
- [ ] The call site uses a `#[cfg(not(unix))]` plain-`spawn()` fallback (matching the
      existing `process_group` cfg split)
- [ ] `retries_past_etxtbsy_until_writer_closes` is `#[cfg(target_os = "linux")]` with a
      comment explaining the macOS/Windows exclusion
- [ ] Full suite + clippy --all-targets + fmt green on all three CI OSes

ssilvius:
Fixes #685.

## Summary

Review follow-up to #683 (the ETXTBSY retry), which was approved with two MED findings that
landed on main. Both are platform-scoping bugs: the retry is POSIX-specific but was compiled
ungated, and its regression test is load-bearing only on Linux.

### 1. ETXTBSY const + retry consts + spawn_with_etxtbsy_retry are #[cfg(unix)]
`ETXTBSY` (a POSIX errno), `ETXTBSY_RETRY_BUDGET`, `ETXTBSY_RETRY_BACKOFF`, and
`spawn_with_etxtbsy_retry` are now `#[cfg(unix)]`. On Windows `raw_os_error()` returns Win32
codes where 26 is `ERROR_SHARING_VIOLATION`, so running the retry there would misidentify a
sharing violation as ETXTBSY and burn the 2s budget -- the gate prevents that path from
existing on non-unix. The `Child` and `Instant` imports those items consume are split to
`#[cfg(unix)]` too, so they do not warn as unused on non-unix (`-D warnings` would fail CI).
Evidence: `src/worksource/mod.rs` `#[cfg(unix)]` on the const block, the helper, and the
`use std::process::Child` / `use std::time::Instant` imports.

### 2. The call site uses a #[cfg(not(unix))] plain-spawn fallback
`call_plugin_inner` selects the spawn path by target: `#[cfg(unix)]` calls
`spawn_with_etxtbsy_retry(&mut cmd)`, `#[cfg(not(unix))]` calls `cmd.spawn()`. Both feed the
same `map_err` so the error message is unchanged. This mirrors the existing `#[cfg(unix)]`
`cmd.process_group(0)` split a few lines above -- the established pattern in this function.
Evidence: `src/worksource/mod.rs` `call_plugin_inner` spawn-path cfg split.

### 3. retries_past_etxtbsy_until_writer_closes is #[cfg(target_os = "linux")] with a comment
The test is gated to Linux with a comment explaining the exclusion: Linux raises ETXTBSY on
`execve` of a shebang script held open for writing; Darwin raises it only for native Mach-O
binaries, so on macOS the held fd never triggers the retry and the first spawn succeeds --
the test would pass vacuously. On Linux the existing `.expect()` is load-bearing: without the
retry the first spawn fails with ETXTBSY and the test panics.
Evidence: `src/worksource/mod.rs` `#[cfg(target_os = "linux")]` + comment on the test.

### 4. Full suite + clippy --all-targets + fmt green on all three CI OSes
Local (macOS host): suite green, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt
-- --check` clean. The Linux retry test and the Windows compile (where the unused-import
warning would have fired before the import split) are confirmed by the CI matrix on this
branch.
Evidence: local runs; the windows-latest + ubuntu-latest + macos-latest CI jobs on this PR.

## Not done

- No change to the retry semantics, budget, backoff, or logging -- only platform scoping.
- No retry-counter threaded out of the helper; the Linux-gated `.expect()` already makes the
  test load-bearing without exposing internals.

### DIFF
--- a/src/worksource/mod.rs
+++ b/src/worksource/mod.rs
@@ -13,10 +13,16 @@ pub use prs::*;
 
 use std::io::Read;
 use std::path::{Path, PathBuf};
-use std::process::{Child, Command, Stdio};
+use std::process::{Command, Stdio};
+// `Child` and `Instant` are used only by the unix-only ETXTBSY retry helper;
+// importing them unconditionally would warn as unused on non-unix targets.
+#[cfg(unix)]
+use std::process::Child;
 use std::sync::{Mutex, OnceLock};
 use std::thread;
-use std::time::{Duration, Instant};
+use std::time::Duration;
+#[cfg(unix)]
+use std::time::Instant;
 
 use wait_timeout::ChildExt;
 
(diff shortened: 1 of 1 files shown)


ssilvius:
Superseded: #683 was reverted (#684); these cfg corrections are folded into the clean re-land PR. Closing in favor of that.

Issue #689, opened 2026-06-27
fix(simplify): require located evidence in legion-simplify articulation


ssilvius:
## Problem

legion-simplify rubber-stamps: it records `issues` on 0.9% of 1131 runs vs legion-pr-write at 21.3% on 254 runs -- same agents, same PRs. The one structural difference: `pr_write::validate_pr_body` requires each mapping entry to cite located evidence (a test, file:line, or symbol) via `has_evidence`; `simplify_check::validate_articulation` enforces only coverage + a per-entry word threshold, with an explicit "No evidence requirement is enforced here (unlike pr-write)" carve-out (src/simplify_check.rs:102-103). Requiring located evidence is the highest-leverage anti-rubber-stamp lever -- the process-supervision / generative-verifier literature and legion's own ~24x within-system natural experiment both point at it.

## Change

Add a located-evidence requirement to `validate_articulation`, reusing `pr_write::has_evidence` (promoted to `pub(crate)` -- no second heuristic). Evidence must come from the entry BODY, not the `### <path>` heading (the heading always contains the file path and would trivially satisfy the source-path check). An entry with sufficient prose but no file:line / symbol / `Evidence:` citation fails validation, so no clean simplify gate is recorded and `legion pr create` refuses until the agent points at the construct it actually examined. Update `plugin/skills/legion-simplify/SKILL.md` to instruct one located citation per file entry.

## Acceptance criteria

- [ ] `validate_articulation` rejects an entry that has sufficient prose but cites no located evidence, with a finding naming the file and asking for a file:line/symbol citation.
- [ ] The evidence check reads the entry body only -- the `### <path>` heading does not satisfy it.
- [ ] `has_evidence` is shared from `pr_write` (promoted to pub(crate)); `simplify_check` reuses it rather than reimplementing.
- [ ] A substantive articulation whose entries cite a symbol or file:line still passes (updated fixture).
- [ ] `plugin/skills/legion-simplify/SKILL.md` documents the per-file located-citation requirement.
- [ ] cargo test, cargo clippy --all-targets -- -D warnings, cargo fmt --check all pass.

## Out of scope

- Phase 2 (gate-trust prediction on the uncertainty engine) -- separate issue.
- legion-review / legion-verify evidence rules -- separate.

ssilvius:
Groom 2026-07-05, closing as shipped via PR #690 (b16ff39): the located-evidence validator is live -- it rejected two under-evidenced articulations in real use this week.

PR #690, opened 2026-06-27
fix(#689): require located evidence in legion-simplify articulation

ssilvius:
## Summary

legion-simplify rubber-stamped: `issues` on 0.9% of 1131 runs vs legion-pr-write's
21.3% on 254 runs, same agents and PRs. The lone structural difference was that
pr-write requires located evidence per entry and simplify did not. This ports that
requirement into the simplify gate, reusing the same detector so the two cannot drift.

### 1. validate_articulation rejects substantive-but-uncited entries
Added a per-entry evidence check that pushes a "cites no located evidence" finding,
naming the file, when `has_evidence(entry_body)` is false -- so a vague clean-claim
with enough prose but no anchor now fails instead of recording clean.
Evidence: src/simplify_check.rs:134; test `fn refuses_entry_without_located_evidence`
asserts !ok, that the finding names src/foo.rs, and that it does not also trip thinness.

### 2. Evidence check reads the body only, not the heading
New `fn entry_body` strips the `### <path>` heading (which always contains the file
path) before the check, so the heading cannot satisfy the source-path test for free.
Evidence: src/simplify_check.rs entry_body; refuses_entry_without_located_evidence
keeps the heading present and still fails, proving the heading does not count.

### 3. Evidence detector shared from pr_write, no parallel heuristic
simplify calls `has_within_file_locator` (strict: symbol / file:line / Evidence:) and
pr-write keeps `has_evidence` (generous: also bare path + behavioral cue, since a PR may
cite observable behavior). Both share the `source_tokens` tokenizer and one `SOURCE_EXTS`
const -- the change removes duplication rather than adding a parallel detector. This split
is the fix for the two pre-push review HIGHs: the original reuse accepted bare cue words,
then a bare filename; the strict locator now rejects both, so the gate enforces the
within-file anchor it advertises (symbols keep clean verdicts ergonomic, no fake line nums).
Evidence: src/pr_write.rs:188 (pub(crate) fn has_within_file_locator) + has_source_path_with_line;
`use crate::pr_write::{..., has_within_file_locator, ...}` in src/simplify_check.rs;
test `fn within_file_locator_excludes_bare_path_and_behavioral`.

### 4. A substantive articulation citing a symbol/file:line still passes
Updated the good_articulation fixture to cite `fn parse_foo` at src/foo.rs:30 and added
accepts_entry_with_explicit_evidence_line for the Evidence: form; both pass.
Evidence: tests `fn accepts_substantive_articulation` and
`fn accepts_entry_with_explicit_evidence_line` in src/simplify_check.rs.

### 5. SKILL.md documents the located-citation requirement
Bumped to version 2.1.0, updated the frontmatter description (coverage + non-boilerplate
+ located evidence), and added a paragraph requiring one located citation per entry.
Evidence: plugin/skills/legion-simplify/SKILL.md:55 and the frontmatter description.

### 6. test / clippy / fmt pass
The full suite stays green after the new check and the four updated fixtures; clippy
runs with --all-targets so test and example code are linted too, and formatting is
unchanged. Ran all three gates locally on this HEAD before opening the PR.
Evidence: 195 tests pass; `cargo clippy --all-targets -- -D warnings` clean;
`cargo fmt -- --check` clean.

## Not done

- Did NOT change legion-review or legion-verify evidence rules (separate scope).
- Did NOT fold the redundant `to_lowercase()` on has_evidence's miss path (two allocs)
  -- a trivial perf nit the pre-commit review flagged as not worth a churn commit; left
  for the next time this function is touched.
- Phase 2 (gate-trust prediction on the uncertainty engine) is issue-tracked separately.

(Note: an earlier draft deferred the behavioral-cue-word soft spot to a follow-up; the
pre-push independent review correctly raised it to HIGH, so it is fixed in this PR -- see
criterion 3 -- not deferred.)

### DIFF
--- a/plugin/skills/legion-simplify/SKILL.md
+++ b/plugin/skills/legion-simplify/SKILL.md
@@ -4,9 +4,9 @@ description: |
   Review the current branch diff for code quality issues: duplicate logic, unnecessary
   abstraction, stringly-typed state, and other structural problems. Articulate, per changed
   file, what you checked and why the verdict holds; `legion quality-gate check` validates the
-  articulation (coverage + non-boilerplate) and records the gate so `legion pr create` can
-  verify clean state before opening a PR. Run this on every feature branch before a PR.
-version: 2.0.0
+  articulation (coverage + non-boilerplate + located evidence) and records the gate so `legion
+  pr create` can verify clean state before opening a PR. Run this on every feature branch before a PR.
+version: 2.1.0
 user-invocable: true
 allowed-tools: Bash, Read
 ---
@@ -52,6 +52,12 @@ reasoning for the verdict. Entries under ~12 words are rejected as boilerplate, 
 "clean" or a restated category list will not pass. A `clean` verdict still requires the
 reasoning -- "I read X, the duplicated-looking Y is justified because Z." A finding states the
 problem, the file:line, and the fix.
+
+Every entry -- clean or finding -- must cite a **located construct** in its body: a `file:line`,
+a symbol (`fn name` / `Type::method`), or an `Evidence:` line. The `### <path>` heading does NOT
+count -- it just restates the file name. Point at the specific line or symbol you actually read.
+An entry whose prose is substantive but cites nothing locatable is refused: "I checked it, looks
+clean" with no anchor is precisely the rubber-stamp this gate exists to stop.
 
 ```markdown
 ### src/db/foo.rs
--- a/plugin/skills/legion-simplify/SKILL.md
+++ b/plugin/skills/legion-simplify/SKILL.md
@@ -53,11 +53,12 @@ reasoning for the verdict. Entries under ~12 words are rejected as boilerplate, 
 reasoning -- "I read X, the duplicated-looking Y is justified because Z." A finding states the
 problem, the file:line, and the fix.
 
-Every entry -- clean or finding -- must cite a **within-file locator** in its body: a `file:line`,
-a symbol (`fn name` / `Type::method`), or an `Evidence:` line. Neither the `### <path>` heading nor
-a bare filename in prose counts -- both just restate the file name. Point at the specific line or
-symbol you actually read. An entry whose prose is substantive but cites nothing locatable is
-refused: "I checked it, looks clean" with no anchor is precisely the rubber-stamp this gate stops.
+Every entry -- clean or finding -- must cite a **within-file locator** in its body: a `file:line`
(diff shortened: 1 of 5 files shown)


ssilvius:
**Rust review (smugglr): APPROVE.**

Reviewed the 5-commit diff (862eae3^..7cad6df): `src/pr_write.rs`, `src/simplify_check.rs`, `src/cli/verify.rs`, fixtures, SKILL.md. Focus per request: idioms + the byte-slice in `has_source_path_with_line`.

**The byte-slice is sound.** `tok[i + needle.len()..]` cannot panic. `match_indices(&needle)` yields byte offsets at the start of an all-ASCII needle (`.rs:` etc.), so `i + needle.len()` always lands on the byte just past the match -- a guaranteed UTF-8 char boundary, even when a multibyte char precedes the match. An exact-trailing-colon token (`bar.ts:`) yields an empty slice, `chars().next()` is `None`, and the entry correctly fails the within-file bar. I extracted the three locator fns standalone and ran 16 cases (including `café.rs:1`, `naïve/path/résumé.py:9`, `bar.ts:`, `foo.rs:42:10`): 0 mismatches, no panic.

**Idioms / design -- clean:**
- The `has_evidence` (generous, pr-write) vs `has_within_file_locator` (strict, simplify) split is the right call, and sharing `source_tokens` + one `SOURCE_EXTS` const means the strict bar can't silently drift from the generous one. This is dedup, not a parallel detector -- matches criterion 3.
- `entry_body` strips only `### ` headings and KEEPS `Evidence:` lines, so the heading can't satisfy the source-path test for free while an explicit citation still short-circuits. The non-ascii-path fixture leaning on an `Evidence:` line (which hits the `contains("evidence:")` early return) sidesteps path-detection entirely -- nicely chosen.
- `has_line_reference` collecting into a `Vec` for `windows(2)` is idiomatic (iterators have no `windows`).
- Tests pin the real risks: bare-path rejected, behavioral-cue rejected, symbol/`file:line`/`Evidence:` accepted, and the evidence failure asserted distinct from the thinness failure.

**Non-blocking nits (do not gate):**
1. `has_within_file_locator`/`has_evidence` allocate `to_lowercase()` solely for the `evidence:` test (has_evidence twice). You already flagged the double-lowercase as not-worth-churn; agreed -- fold it next time the fn is touched.
2. `has_source_path_with_line` re-runs `format!("{ext}:")` per ext per token. Articulations are short; immaterial.

No correctness, security, or silent-failure issues. The pre-push HIGH (bare cue / bare filename) is genuinely closed by the strict locator, with regression guards. Ship it.

Issue #714, opened 2026-07-01
fix: legion sub-issue list fails with 'missing field labels' deserialization error


ssilvius:
## Goal

**`legion sub-issue list` returns results instead of failing to deserialize the GitHub response.**

### Behavior
- `legion sub-issue list --repo legion --number 704` (or equivalent) currently fails with
  `missing field labels` -- a serde deserialization error on the sub-issue listing path. The
  create/link path works (epic #704's children #705-#713 were linked successfully with it).
- Fix the response struct: the field is either absent from GitHub's sub-issue API payload or
  optional -- model it as `Option<...>`/`#[serde(default)]` rather than required.
- Add a test deserializing a captured/representative API payload without the field.

### Error Handling
- Deserialization failures should name the endpoint and the offending field (they did here --
  keep that property).

## Out of Scope
- Any sub-issue create/link changes (working).

## File Locations
- Implementation: the sub-issue work-source client (where the list response struct lives)
- Tests: `#[cfg(test)]` same file

## Done When
- [ ] `legion sub-issue list` returns the children of #704
- [ ] Payload-without-labels test passes
- [ ] Tests pass; simplify + review done; PR linked

**This issue is complete when:** listing an epic's children works against the live GitHub API.

## Context
- Observed 2026-07-01 while linking #705-#713 to epic #704 (checkpoint 019f1f6f).

Issue #713, opened 2026-07-01
feat: fair-guard fixes -- reword messages, lane-4 stance, coverage guarantee


ssilvius:
## Goal

**Make the grep/find guard FAIR and stop teaching agents the wrong mental model: reword the guard messages, fix the `legion-explore` lane-4 framing, and GUARANTEE per-language index coverage so "sym is the universal index over all files/languages" is both TRUE and TAUGHT.**

### Behavior
- **Message rewrite** (`plugin/hooks/pre-bash-grep.sh`, `plugin/hooks/pre-grep.sh`): remove wording
  that implies sym is rust-only or symbols-only; route non-symbol/literal/file-discovery queries to
  `legion sym etc` (find-content / extract / tree / find-file) instead of presenting shell grep as
  the escape. The block message must print the exact runnable command shape for each query type --
  agents use what the guard prints. Keep the soft-bypass path (`pre-bash-grep.sh:119-121`) only for
  genuinely uncovered cases.
- **Lane-4 framing** (the `legion:legion-explore` agent definition): rewrite so bypass is LAST
  resort -- try `sym`, then `sym etc`, then Grep tool; a bypass is a signal the index is missing an
  answer, logged, not a normal move. Kill the "empty sym => bypass is fine" default.
- **Coverage guarantee** (`src/scip.rs` + `_legion-indexed.sh`): every language `detect_languages`
  reports for a repo must actually get indexed (or the gap is surfaced loudly), so a correct "sym
  does TS" belief is not undercut by a silently-missing TS index.

### Metric design (revised at plan review 2026-07-01)
- This issue rewords the guard, which changes what gets CLASSIFIED as a bypass mid-experiment --
  so raw before/after bypass counts are the SECONDARY signal only.
- PRIMARY signal: `sym etc` / `sym tree` usage counts and zero-result rate, instrumented in
  E2/E3/E4 from day one (`src/telemetry.rs`). The root cause was "tools return nothing"; the
  metric that proves the fix is "agents use the tools and they return something."
- Report both in the epic close-out: usage + zero-result rate trend, and per-shape bypass counts.

### Open scope decision (operator call, blocking the coverage half)
- Whether fixing #442 (daemon dispatch: watched TS repos never auto-index until a manual
  `legion index <repo>`) is IN scope for the coverage guarantee. If in, this issue grows a real
  daemon change; if out, the guarantee degrades to "loudly surfaced" rather than "actually
[...]

Issue #705, opened 2026-07-01
feat: file-inventory table populated by the index-build walk


ssilvius:
## Goal

**Persist a complete, structured file inventory for every watched repo during `legion index` -- the always-fresh substrate that `sym tree` and `sym etc find-file` query.**

### Interface
```rust
// src/db/inventory.rs (new) -- or fold into src/db/scip.rs
pub struct FileInventoryEntry {
    pub repo: String,
    pub path: String,        // repo-relative
    pub ext: Option<String>,
    pub lang: Option<String>,// SCIP language name, None for non-indexed types
    pub size: u64,
    pub mtime: String,       // rfc3339
    pub symbol_count: u32,   // 0 when no SCIP symbols for this file
}

impl Database {
    pub fn upsert_file_inventory(&self, entries: &[FileInventoryEntry]) -> Result<()>;
    pub fn list_file_inventory(&self, filter: &InventoryFilter<'_>) -> Result<Vec<FileInventoryEntry>>;
    pub fn prune_file_inventory(&self, repo: &str, live_paths: &[String]) -> Result<usize>;
}
```

### Behavior
- A NEW recursive walk enumerates every non-ignored file in the repo and records one inventory
  row per file. There is no existing walk to extend: `scip::run_indexer` (`src/scip.rs`) only
  dispatches to external SCIP indexer binaries, and `detect_languages` is root-marker sniffing
  that never recurses. Use the `ignore` crate for the walk (gitignore-aware; note `walkdir` is
  neither a current dep nor gitignore-aware). Invoke the walk from the `legion index` path
  (`handle_index`, `src/cli/index_cmd.rs`) alongside the SCIP loop.
- GATE RELAXATION: `handle_index` currently hard-errors when `detect_languages` returns empty
  (`src/cli/index_cmd.rs:950`). Relax it -- inventory population must run for docs-only repos
  (zero SCIP languages); skip the SCIP loop with an info message instead of erroring. A
  pure-markdown/astro-content repo is a headline use case for this epic.
- Per-file `lang` comes from a NEW ext-to-lang mapping (e.g. `lang_for_ext`), NOT from
  `detect_languages` -- that function is repo-level marker detection and cannot classify
  individual files. Non-code files land with `lang = None`. `symbol_count` from the SCIP blob
  where one exists, else 0.
- Upsert keyed on `(repo, path)` -- idempotent re-index. After the walk, `prune_file_inventory`
  removes rows for files no longer present (deletes/renames).
[...]

Issue #704, opened 2026-07-01
feat: legion sym etc -- comprehensive non-symbol file index + fair guard (epic)


ssilvius:
## Goal

One comprehensive, structured, always-fresh answer surface over **every file `sym` does not parse**
(docs, configs, css, astro, mdx, json, yaml, prose, lockfiles), returning STRUCTURED data, so
agents never need `cat`/`grep`/`find` to orient or extract on non-symbol files -- and the
grep/find enforcement guard becomes FAIR because there is an answer for every query shape.

This is a tracking epic. Implementation lands in the linked child issues.

## Why (evidence)

Bypass telemetry: 87 real Bash escapes of the grep guard in 30 days, 79 of them from
`legion-explore` (the agent built to use `sym`), and `had_sym`/`had_recall` fired on ~0 of them
-- agents bypass precisely when the tools return nothing. The gaps are structural: `sym` (SCIP)
answers only symbol questions, while agents need string-literal search, config/frontmatter
extraction, file/layout discovery, and cross-language answers. Agents also prefer python over
`Glob` because they want STRUCTURED data, not a flat path list.

## Decomposition (child issues)

Four query shapes (bypass-volume ranked) plus the engine and the guard fix:

- **E1 #705** File-inventory table, populated by a new gitignore-aware walk (substrate for tree/find-file).
- **E2 #706** `sym tree` -- build-tree shape (structured layout), instrumented from day one.
- **E3 #707** `sym etc find-content` -- exact content search on the ripgrep engine crates
  (direct disk scan: grep parity incl. regex, fresh by construction, no corpus). Biggest bypass class.
- **E4 #708** `sym etc extract` -- config/frontmatter bytes via jq-style dotted path, without full read.
- **E5 #709** `sym etc find-file` -- locate files by name/role across repos (inventory-backed).
- **E6 #710** Module-graph engine (`oxc_parser` + `oxc_resolver`) -- "which files import X".
- **E7 #711** CSS symbols via `lightningcss` (class + custom-property defs).
- **E8 #712** Astro handling (frontmatter + script) -- deferrable.
- **E9 #713** Fair-guard fixes -- reword the guard messages + `legion-explore` lane-4 framing, and
  guarantee per-language index coverage (open scope call: #442 in or out).

## Engine decisions (settled, with evidence)

- **find-content = ripgrep engine crates** (`grep-searcher`/`grep-regex`/`ignore`), NOT Tantivy
[...]

PR #717, opened 2026-07-02
fix(#714): serde default on ExternalIssue.labels so sub-issue list deserializes

ssilvius:
## Summary

`legion sub-issue list` failed on every call with `missing field labels`: the GitHub sub-issue GraphQL API returns only `{number, url, title, state, body}` per child, and `ExternalIssue.labels` was a required `Vec<serde_json::Value>`, so serde refused the whole payload before any row could print. This change adds `#[serde(default)]` to the field so an absent `labels` deserializes to an empty vec -- the correct semantic for an API that simply does not send the field -- without changing the field's type or disturbing the `gh issue list` path, which does supply labels and keeps deserializing exactly as before.

### 1. `legion sub-issue list` returns the children of #704

The failure was in deserialization, not the query: the GraphQL call succeeded and the payload arrived, but serde rejected it because `labels` was modeled as required while the sub-issue endpoint omits it. With `#[serde(default)]` on `ExternalIssue.labels` (src/worksource/issues.rs:21), the absent field materializes as an empty vec and the rows flow through to output. Verified live against GitHub: `legion sub-issue list --repo legion --parent 704` now prints all nine children, #705 through #713, with their titles and OPEN state.
Evidence: live run output listing #705-#713; src/worksource/issues.rs:21

### 2. Payload-without-labels test passes

`external_issue_deserializes_without_labels_field` pins a representative sub-issue payload containing exactly the five fields the endpoint sends (`number`, `url`, `title`, `state`, `body`) and asserts the struct populates with `labels` empty, `assignees` and `created_at` None -- the regression guard that fails if the default is ever removed. A companion test, `external_issue_deserializes_with_labels_field`, proves the other direction: a payload that does carry `labels` (the `gh issue list` shape) still round-trips with the entries intact, so the fix cannot silently break the existing path.
Evidence: fn external_issue_deserializes_without_labels_field, fn external_issue_deserializes_with_labels_field in src/worksource/issues.rs

### 3. Tests pass; simplify + review done; PR linked

The full worksource test suite passes on the branch (41 tests), and `cargo clippy --all-targets -- -D warnings` plus `cargo fmt -- --check` are both clean. The simplify gate is recorded clean on HEAD (gate 019f208d-0015), this body is the pr-write gate, and the PR is opened via `legion pr create` referencing #714 so the link is structural. Review follows via /legion-review and team review from vault and smugglr.
Evidence: quality-gate row 019f208d-0015-75a2-aa7f-74ba01d922eb on commit 60912ea; cargo test output 41 passed

## Not done

The `labels` field stays `Vec<serde_json::Value>` rather than a typed label struct -- tightening it is orthogonal to this fix and would touch the working `gh issue list` path for no behavioral gain. Error-message quality on deserialization failures (naming endpoint and field) is untouched: it already had that property, and nothing here removes it. No changes to the sub-issue create/link path, which was explicitly out of scope and already working.

### DIFF
--- a/src/worksource/issues.rs
+++ b/src/worksource/issues.rs
@@ -16,6 +16,11 @@ pub struct ExternalIssue {
     pub number: u64,
     pub title: String,
     pub body: Option<String>,
+    /// Label list. The GitHub sub-issue GraphQL API omits this field entirely
+    /// (it returns only `number`, `url`, `title`, `state`, `body`), so we
+    /// default to an empty vec rather than failing deserialization when the
+    /// field is absent. (#714)
+    #[serde(default)]
     pub labels: Vec<serde_json::Value>,
     pub assignees: Option<Vec<serde_json::Value>>,
     pub state: String,
@@ -469,5 +474,60 @@ mod tests {
             msg.contains("at least one of title or body"),
             "expected validation error message, got: {msg}"
         );
+    }
+
+    /// Regression guard for #714: the GitHub sub-issue GraphQL API returns
+    /// `{ number, url, title, state, body }` -- it omits `labels` entirely.
+    ///
+    /// Without `#[serde(default)]`, serde fails with "missing field `labels`".
+    /// This test deserializes a representative sub-issue payload (no `labels`,
+    /// no `assignees`, no `createdAt`) and asserts that the struct populates
+    /// with sensible defaults rather than erroring.
+    #[test]
+    fn external_issue_deserializes_without_labels_field() {
+        let json = r#"{
+            "number": 705,
+            "url": "https://github.com/runlegion/legion/issues/705",
+            "title": "E1: sym index inventory",
+            "state": "OPEN",
+            "body": "Implement the E1 sub-task."
+        }"#;
+
+        let issue: ExternalIssue =
+            serde_json::from_str(json).expect("should deserialize without labels field");
+
+        assert_eq!(issue.number, 705);
+        assert_eq!(issue.url, "https://github.com/runlegion/legion/issues/705");
+        assert_eq!(issue.title, "E1: sym index inventory");
+        assert_eq!(issue.state, "OPEN");
+        assert!(
+            issue.labels.is_empty(),
+            "labels should default to empty vec, got {:?}",
+            issue.labels
+        );
+        assert!(issue.assignees.is_none());
+        assert!(issue.created_at.is_none());
+    }
+
+    /// Verify that deserialization also works when `labels` IS present (the
+    /// existing `gh issue list` path still works after the default is added).
+    #[test]
+    fn external_issue_deserializes_with_labels_field() {
+        let json = r#"{
+            "number": 42,
+            "url": "https://github.com/runlegion/legion/issues/42",
+            "title": "Some issue",
+            "state": "open",
+            "body": null,
+            "labels": [{"id": "LA_1", "name": "bug"}, {"id": "LA_2", "name": "critical"}],
+            "assignees": [],
+            "createdAt": "2026-06-01T00:00:00Z"
+        }"#;
+
+        let issue: ExternalIssue =
+            serde_json::from_str(json).expect("should deserialize with labels field");
+
+        assert_eq!(issue.labels.len(), 2);
+        assert_eq!(issue.created_at.as_deref(), Some("2026-06-01T00:00:00Z"));
     }
 }


ssilvius (commented):
/legion-review result: approved (gate 019f2094-e78b, clean on HEAD 60912ea). Combined single-agent review (PR under 100 lines) across spec, correctness, quality, security; 1 finding, adversarially verified, 0 refuted.

Surviving MED (mitigated toward LOW, no code change required): the PR body's "Not done" section claims deserialization failures already name the endpoint -- they do not. The actual error string is "work source error: missing field `labels` at line 1 column N" (src/worksource/mod.rs:384, src/error.rs:70): serde names the field, nothing names the endpoint. The diff does not touch that path, so nothing regressed and #714's "keep that property" is vacuously met, but the claim itself is wrong. Correction recorded here since PR bodies are not editable via the sanctioned surface; follow-up filed as #720 to thread operation context through plugin_call's deserialization errors.

Spec: all three acceptance criteria verified, including a live run returning #705-#713. Correctness: serde(default) is the right mechanism; both payload shapes pinned by tests. Quality: gates clean (simplify 019f208d-0015, pr-write, review). Security: no surface.

ssilvius:
Rust-lane review (smugglr), requested via signal 019f2096-0bb6. Verdict: correct and minimal -- APPROVE (posted as comment; shared identity cannot formally approve).

What I verified, not just read:

1. Premise checked against the plugin, not the PR comment. plugin/worksources/github list-sub-issues GraphQL query selects exactly `number url title state body` -- so `labels` is genuinely absent on that path. Of the remaining fields, `assignees` is Option (serde implicit None on missing) and `created_at` already had #[serde(default, alias = "createdAt")]. `labels: Vec<_>` was the only rigid non-required field; #[serde(default)] is exactly the right scope of fix.

2. list_sub_issues (src/worksource/issues.rs:207) deserializes plugin stdout straight into Vec<ExternalIssue> via plugin_call, so the fix covers the failing path directly.

3. Consumer safety: the only reader of ExternalIssue.labels is the import path (src/worksource/issues.rs:294-317), which already branches on empty -> labels None, priority Med. No consumer distinguishes missing-vs-empty.

4. No wire change on output: serde(default) affects deserialization only; Serialize still always emits "labels": [] -- identical to today's populated-path output shape, so `issue view`/JSON consumers are unaffected.

5. Ran both new tests at head 60912ea in a clean worktree: 2 passed. The pair is well-chosen -- one pins the GraphQL sub-issue shape (uppercase OPEN, no labels/assignees/createdAt), the other pins the gh-issue-list shape (lowercase state, object labels, createdAt alias) so the default cannot regress the existing list path.

Alternative considered and rejected in agreement with the PR: adding `labels { nodes { name } }` to the GraphQL query would return a different shape ({nodes:[...]}) than the gh --json shape ([{id,name}]) the label-extraction code expects, and tolerant deserialization also hardens against any third-party worksource plugin omitting the field. The Rust-side default is the more robust fix.

Nothing blocking. Pre-existing looseness (labels as Vec<serde_json::Value>) is out of scope, and the accepted nit about deser errors not naming the endpoint is already filed as #720.

ssilvius (commented):
Rust-lane team review (smugglr). Verdict: APPROVE (posted as comment -- GitHub refuses same-account approvals).

What I checked, beyond reading the diff:

1. Field semantics: swept every consumer of ExternalIssue.labels via sym refs. The only real consumer is sync_issues_to_kanban (src/worksource/issues.rs:294), which maps label objects to card labels and priority. An absent field defaulting to empty vec yields labels=None, priority=Med -- correct for a payload from an endpoint that has no label concept. That path is fed by the plugin 'list' call (gh issue list shape), which always supplies labels, so no behavior change there. Serialize side unaffected: --json output emits labels: [], which is honest.

2. Struct audit: confirmed labels was the only required field the sub-issue payload omits -- assignees and created_at are already Option (created_at also has serde(default, alias)), so no second latent 'missing field' bug is hiding behind this one.

3. Tests: ran the branch locally -- 41 worksource tests pass, and both new deser tests pass in the lib target. Mutation check: removed the serde(default) attribute and re-ran; external_issue_deserializes_without_labels_field fails as intended, so the regression guard is load-bearing, not decorative. The with-labels companion pins the gh-issue-list shape against silent breakage in the other direction.

4. Fail-open note (accepted, not a blocker): a buggy plugin that wrongly omits labels now deserializes silently instead of erroring. Given the field is presentation-only (labels/priority hints) and the API genuinely omits it, absent==empty is the right call.

CI green across all seven checks. Ship it.

Issue #709, opened 2026-07-01
feat: sym etc find-file -- locate files by name/role across repos


ssilvius:
## Goal

`legion sym etc find-file` -- locate files by name or role across all watched repos, from the
inventory (the find-file shape, 20+ bypasses: "locate the huttspawn checkout", "which repo owns X").

### Interface
```rust
// legion sym etc find-file <name-or-glob> [--repo] [--role <config|test|doc|entry>] [--json]
pub fn find_file(query: &str, repo: Option<&str>, role: Option<Role>) -> Result<Vec<FileInventoryEntry>>;
```

### Behavior
- Queries `Database::list_file_inventory` (E1) by basename / glob across repos (cross-repo default).
- Optional `--role` maps to heuristics over path/name/ext (config, test, doc, entrypoint).
- `--json` structured; each hit carries its repo + path so cross-repo answers are unambiguous.

### Error Handling
- No match => explicit empty-with-hint, not a bare empty list on stdout.

## Out of Scope
- Content search (E3). May reuse E2's query layer rather than duplicate it.

## File Locations
- Implementation: `src/cli/index_cmd.rs` + `src/db/inventory.rs`
- Tests: `#[cfg(test)]` in `src/db/inventory.rs`

## Dev Workflow
1. Build (tests; cargo test / clippy -D warnings / fmt --check). 2. `/simplify`. 3. `/review-pr`. 4. Fix. 5. PR referencing this issue.

### Rust Rules
No `unwrap()` in prod, no `unsafe`, no emoji, clippy -D warnings clean, fmt clean, thiserror errors.

## Done When
- [ ] `legion sym etc find-file components.json` finds it across repos, tagged by repo
- [ ] `--role config` surfaces config files; `--repo` scopes
- [ ] Tests pass; simplify + review done; PR linked

**Complete when:** an agent can find a file by name/role across repos without `find` or knowing the checkout path.

## Context
- Epic: #704. Depends on #705 (inventory). Design: 019eadca.

Issue #706, opened 2026-07-01
feat: sym tree -- structured cross-repo build-tree query


ssilvius:
## Goal

**`legion sym tree` -- a STRUCTURED, cross-repo view of the file inventory (the build-tree shape). The sanctioned replacement for `find` / `ls -R` / `tree` / throwaway `os.walk` python.**

Structured JSON is the point: agents reach for python because `Glob` returns a flat path list, not objects.

### Interface
```rust
// src/cli/index_cmd.rs -- new SymAction variant
Tree {
    #[arg(long)] repo: Option<String>,   // omitted = all watched repos (watch.toml)
    #[arg(long)] ext: Option<String>,    // filter by extension
    #[arg(long)] under: Option<String>,  // subtree prefix
    #[arg(long)] depth: Option<u32>,     // max depth
    #[arg(long)] json: bool,
}
```

### Behavior
- Queries `Database::list_file_inventory` (E1 #705) -- no filesystem walk at query time.
- `--repo` omitted means cross-repo over every repo in `watch.toml`; each entry carries its repo.
- `--ext` / `--under` / `--depth` filter server-side so the agent gets exactly the slice.
- `--json` emits structured entries `{repo, path, ext, lang, size, symbol_count}`; default is a
  compact human table (mirrors `run_sym_list`'s output style, `src/cli/index_cmd.rs:157`).
- Empty/uncovered repo produces an explicit "no inventory; run `legion index <repo>`" message,
  not silence.
- Telemetry: record every invocation (filters used, result count) via `src/telemetry.rs`.
  Zero-result rate is the epic's PRIMARY success metric, instrumented from day one (moved up
  from E9 -- raw bypass counts alone cannot distinguish "agents do not use it" from "they use
  it and it fails them").

### Error Handling
- Unknown repo / not in watch.toml returns a clear `LegionError` with the fix hint.

## Out of Scope

- Full-text content (E3 #707) and by-name/role find-file (E5 #709, may reuse this query layer).

## File Locations

- Implementation: `src/cli/index_cmd.rs` (SymAction variant + handler), query via `src/db/inventory.rs`
- Tests: `#[cfg(test)]` in `src/db/inventory.rs` / handler tests

## Dev Workflow

Each step is mandatory. Do not skip steps or combine them.

1. **Build** -- Implement the feature. Write tests alongside code. Run `cargo test`, `cargo clippy -- -D warnings`, `cargo fmt -- --check`. All must pass.
[...]

PR #718, opened 2026-07-02
feat(#705): file-inventory table populated by a gitignore-aware walk on legion index

### LINKED ISSUE #705: feat: file-inventory table populated by the index-build walk

ssilvius:
## Summary

Implements #705: a `file_inventory` table populated on every `legion index` run -- one row
per non-ignored file in the watched repo, pruned when files disappear. This is the always-
fresh structured substrate that `sym tree` (#706) and `sym etc find-file` (#709) will query.
New `src/inventory.rs` (gitignore-aware walk + ext-to-lang mapping) and `src/db/inventory.rs`
(DDL + upsert/list/prune), wired into `handle_index` after -- and independent of -- the SCIP
loop, which docs-only repos now skip instead of erroring.

### 1. Inventory populated for a real repo on legion index, one row per non-ignored file

`inventory::walk_repo` drives an `ignore::WalkBuilder` over the workdir (require_git(false),
hidden(false), explicit `.git` filter_entry) and returns one FileInventoryEntry per file;
`handle_index` upserts the batch in one transaction and logs the count. Dotfiles like
`.github/workflows/ci.yml` are included -- the pre-commit review caught that WalkBuilder's
hidden(true) default silently excluded them, fixed with a pinning test (dotfiles_are_included).
Evidence: tests walk_produces_one_row_per_file, dotfiles_are_included,
gitignored_files_are_excluded (src/inventory.rs).

### 2. Non-SCIP files present with lang = None, per-file lang from ext mapping

Per-file `lang` comes from the new `lang_for_ext` (src/inventory.rs:22), a flat ext-to-SCIP-tag
map -- NOT from `detect_languages`, which is repo-level marker sniffing and cannot classify
individual files. `.sh`/`.md`/`.toml`/`.yaml`/`.css` all return None and land as inventory rows
with lang NULL.
Evidence: tests sh_and_md_have_no_lang, lang_for_ext_unknown_returns_none.

### 3. Docs-only repo gets a populated inventory instead of an error

`handle_index` previously hard-errored when `detect_languages` returned empty; the guard is now
an eprintln + skip of the SCIP block only, and the inventory walk runs unconditionally after it.
The "markers present but every indexer failed" case still hard-errors inside the else arm, so
the relaxation does not swallow real indexing failures.
Evidence: test docs_only_repo_produces_inventory; the restructured guard in
src/cli/index_cmd.rs handle_index.

### 4. Re-index is idempotent; deleted files are pruned

Upsert is keyed ON CONFLICT(repo, path) DO UPDATE, so re-running mutates metadata in place.
After each walk, `prune_file_inventory` deletes rows whose path is absent from the live set --
via a temp table rather than NOT IN (?,...), specifically to dodge SQLite's bound-variable
limit on large repos. During review, the branch's original claim that the ignore crate skips
`.git/` by itself was empirically DISPROVEN (a regression test showed .git/objects entering
the walk once hidden(false) is set); fixed with an explicit filter_entry in commit
"fix(#705): exclude .git explicitly", test git_dir_is_excluded_even_with_hidden_false.
Evidence: db tests in src/db/inventory.rs covering upsert-idempotence and prune; test
git_dir_is_excluded_even_with_hidden_false.

### 5. Tests pass; simplify + review done; PR linked

1175 unit tests pass (18 in inventory modules) with clippy -D warnings and fmt clean. The 4
failing integration tests in tests/integration/worksource_pr.rs are the pre-existing 1Password
signing environment issue, identical on main. Simplify gate recorded clean on HEAD
(019f2088-983b, 1 finding: the .git exclusion, fixed in-branch).

## Not done

- `symbol_count` population from SCIP blobs -- deferred to E6/E7 per the issue's Out of Scope;
  every row lands with symbol_count = 0.
- The read API (`list_file_inventory` / `InventoryFilter`) ships #[allow(dead_code)]-gated:
  its consumers are #706 (sym tree) and #709 (find-file), which remove the gate.
- No incremental single-file inventory update on `legion index --file` -- the full walk is
  cheap (one readdir pass) and runs on the existing repo-level path; per-file incrementalism
  can follow if telemetry shows index latency pain.

Closes #705.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -260,6 +260,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.12.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5cee35f73844aa3014bb606320a6c1f010249dbdf43342fe54b5a4f6a8ed4b79"
+dependencies = [
+ "memchr",
+ "serde_core",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.20.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -978,6 +988,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "globset"
+version = "0.4.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "52dfc19153a48bde0cbd630453615c8151bce3a5adfac7a0aebfbf0a1e1f57e3"
+dependencies = [
+ "aho-corasick",
+ "bstr",
+ "log",
+ "regex-automata",
+ "regex-syntax",
+]
+
+[[package]]
 name = "h2"
 version = "0.4.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1357,6 +1380,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "ignore"
+version = "0.4.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe112b004901c62c2faa11f4f75e9864e0cc5af8da71c9115d184a3aa888749f"
+dependencies = [
+ "crossbeam-deque",
+ "globset",
+ "log",
+ "memchr",
+ "regex-automata",
+ "same-file",
+ "walkdir",
+ "winapi-util",
+]
+
+[[package]]
 name = "indexmap"
 version = "2.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1483,6 +1522,7 @@ dependencies = [
  "dirs 5.0.1",
  "hex",
  "hostname",
+ "ignore",
  "libc",
  "mime_guess",
  "model2vec-rs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,10 @@ hostname = "0.4"
 # on Windows.
 portable-pty = "0.9"
 
+# File inventory walk (#705): gitignore-aware recursive walk over watched
+# workdirs. The same crate powers the ripgrep engine used by find-content.
+ignore = "0.4"
+
 # libc pulls in only on unix for killpg in call_plugin_inner's timeout arm.
 # Must stay AFTER all platform-independent deps -- TOML tables consume
 # every key until the next header, so listing it mid-file would leak the
--- a/src/cli/index_cmd.rs
+++ b/src/cli/index_cmd.rs
@@ -967,7 +967,7 @@ pub(crate) fn handle_index(
     // independent of SCIP outcomes -- including the all-indexers-failed hard
     // error below, which must not leave the inventory stale (#705 review).
     let outcome = inventory::walk_repo(&repo, &repo_path);
-    let live_paths: Vec<String> = outcome.entries.iter().map(|e| e.path.clone()).collect();
+    let live_paths: Vec<&str> = outcome.entries.iter().map(|e| e.path.as_str()).collect();
     database.upsert_file_inventory(&outcome.entries)?;
     // Prune only on a complete walk: with walk errors the entry set may be
     // missing files that still exist, and evicting their rows would let a
--- a/src/cli/index_cmd.rs
+++ b/src/cli/index_cmd.rs
@@ -946,6 +946,19 @@ pub(crate) fn handle_index(
     };
     let repo_path = std::path::PathBuf::from(&entry.workdir);
 
+    // A missing workdir must fail BEFORE the walk: walk_repo on a vanished
+    // root returns an empty entry set, and pruning against it would delete
+    // every inventory row for the repo -- a transient mount failure (watched
+    // workdirs live on external volumes) silently destroying derived state
+    // (#718 review).
+    if !repo_path.is_dir() {
+        return Err(error::LegionError::WatchConfig(format!(
+            "workdir {} for repo '{repo}' does not exist or is not a directory -- \
+             refusing to index (an unmounted volume must not wipe the inventory)",
+            repo_path.display()
+        )));
+    }
+
     let langs = scip::detect_languages(&repo_path);
     let database = open_db()?;
 
@@ -974,6 +987,8 @@ pub(crate) fn handle_index(
         );
     } else {
         let mut indexed: u32 = 0;
+        let mut symbol_counts: std::collections::HashMap<String, u32> =
+            std::collections::HashMap::new();
         for lang in &langs {
             let blob = match scip::run_indexer(lang, &repo_path) {
                 Ok(b) => b,
@@ -998,12 +1013,28 @@ pub(crate) fn handle_index(
             database.upsert_scip_index(&index)?;
             eprintln!("[legion] indexed {repo} ({lang}): {bytes_len} bytes, hash {hash_prefix}");
             indexed += 1;
+            // Per-file symbol counts for the inventory (#705): a blob this
+            // run just built failing to parse is not fatal to indexing, but
+            // it must be loud -- counts silently stuck at 0 would misreport
+            // every SCIP-covered file.
+            match sym::symbol_counts_per_file(&index.blob) {
+                Ok(counts) => symbol_counts.extend(counts),
+                Err(e) => {
+                    eprintln!(
+                        "[legion] per-file symbol counts unavailable for {repo} ({lang}): {e}"
+                    )
+                }
+            }
         }
         if indexed == 0 {
             return Err(error::LegionError::WatchConfig(format!(
                 "no language indexed for {repo} -- every detected language ({}) failed; see warnings above",
                 langs.join(", ")
             )));
+        }
+        if !symbol_counts.is_empty() {
+            let updated = database.update_file_symbol_counts(&repo, &symbol_counts)?;
+            eprintln!("[legion] symbol counts applied to {updated} inventoried files for {repo}");
         }
     }
 
--- a/src/cli/index_cmd.rs
+++ b/src/cli/index_cmd.rs
@@ -998,6 +998,7 @@ pub(crate) fn handle_index(
         );
     } else {
         let mut indexed: u32 = 0;
+        let mut counts_available = false;
         let mut symbol_counts: std::collections::HashMap<String, u32> =
             std::collections::HashMap::new();
         for lang in &langs {
@@ -1029,7 +1030,10 @@ pub(crate) fn handle_index(
             // it must be loud -- counts silently stuck at 0 would misreport
             // every SCIP-covered file.
             match sym::symbol_counts_per_file(&index.blob) {
-                Ok(counts) => symbol_counts.extend(counts),
+                Ok(counts) => {
+                    counts_available = true;
+                    symbol_counts.extend(counts);
+                }
                 Err(e) => {
                     eprintln!(
                         "[legion] per-file symbol counts unavailable for {repo} ({lang}): {e}"
@@ -1043,7 +1047,12 @@ pub(crate) fn handle_index(
                 langs.join(", ")
             )));
         }
-        if !symbol_counts.is_empty() {
+        // Gate on parse success, not on non-empty counts: a blob that parsed
+        // but yielded zero definitions (the repo's last definition was
+        // deleted) must still run the enrich pass so its reset clears the
+        // now-stale counts. A parse FAILURE skips the pass entirely,
+        // preserving prior enrichment.
+        if counts_available {
             let updated = database.update_file_symbol_counts(&repo, &symbol_counts)?;
             eprintln!("[legion] symbol counts applied to {updated} inventoried files for {repo}");
         }
--- a/src/cli/index_cmd.rs
+++ b/src/cli/index_cmd.rs
@@ -7,7 +7,7 @@ use clap::Subcommand;
 
 use crate::cli::datadir::data_dir;
 use crate::cli::util::{open_db, open_db_and_index};
-use crate::{db, error, scip, sym, watch};
+use crate::{db, error, inventory, scip, sym, watch};
 
 #[derive(Subcommand)]
 pub(crate) enum SymAction {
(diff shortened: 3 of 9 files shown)


ssilvius (commented):
/legion-review result: changes_requested (gate 019f20a5-842c on HEAD 12abd7e; posted as comment -- shared identity blocks formal request-changes on own PR). Four dimensions fanned out (spec, correctness, quality, security); 9 candidate findings, each adversarially verified; 2 refuted, 7 survive (3 HIGH, 1 MED, 4 LOW).

HIGH 1 -- symbol_count spec gap (src/inventory.rs:100). Issue #705's Behavior section requires "symbol_count from the SCIP blob where one exists, else 0". Every row is written 0 unconditionally, and the Out-of-Scope bullet the PR body cites covers only NON-SCIP formats (E6/E7 = oxc/lightningcss); .rs/.ts/.py files whose blob was just built in the same handle_index call are inside the spec. Fix: extract per-file symbol counts from the freshly built blob(s), or get the issue re-scoped before merge.

HIGH 2 -- missing workdir wipes the inventory (src/inventory.rs:63-79 + src/db/inventory.rs:168-175). A missing or unmounted workdir (they live on /Volumes mounts) makes walk_repo log the root error and return an EMPTY vec; handle_index then feeds empty live_paths to prune, whose empty branch deletes every row for the repo. Empirically confirmed. Fix: guard repo_path existence in handle_index before the walk (loud WatchConfig error), so a transient mount failure cannot silently destroy derived state.

HIGH 3 -- doc contradicts code (src/inventory.rs:59-62 vs 100-103). The walk_repo doc says the mtime fallback to now() "is logged at the site in that event"; the unwrap_or_else has no logging. Add the log or fix the doc.

MED -- the docs-only gate relaxation (the headline behavior change) has no CLI-level test. tests/integration/index_telemetry.rs already contains a full legion-index harness (watch.toml + fixture repo); a markerless-fixture variant asserting the new stderr lines covers it.

LOW (fix-if-cheap, non-blocking): sh_and_md_have_no_lang lacks a len assert (vacuous if the walk regresses); no symlink-exclusion pinning test; upsert+prune non-atomic (self-heals next run); to_string_lossy PK collision on non-UTF-8 names (macOS APFS makes this near-impossible; precedent at src/watch/config.rs:446).

Refuted (for the record): large-N prune test (the temp-table DELETE binds one param regardless of N -- structurally cannot hit the 999 limit); symlink drop as a correctness defect (correct-by-design, inline-documented).

Fixes are being applied on the branch; re-review will re-gate the new HEAD.

ssilvius (commented):
## Rust-lane review (smugglr) -- reviewed at local tip 0bdfcb7

Note: remote head is 12abd7e; this review covers the three newer local commits too (5e159e4, f8e44e3, 0bdfcb7 -- the docs-only CLI test, inventory-before-SCIP, symbol counts + missing-workdir guard). Verified locally on 0bdfcb7: full suite green (1177 lib + 197 integration), `cargo clippy --all-targets` clean. `unchecked_transaction` in the new db module matches the established db-layer convention (12 existing call sites).

### MED -- fix before merge: an unreadable workdir root still wipes the inventory

Empirically reproduced against the 0bdfcb7 binary: a workdir that **exists but cannot be read** (chmod 000 on the root -- the shape of a permission-mangled external-volume remount, or a Linux-style mount point that is present but empty/unreadable) passes the new `is_dir()` guard, `walk_repo` eprintlns one walk error and returns an empty vec, and the prune pass deletes every row:

```
[legion] inventory walk error (skipped): ...: Permission denied (os error 13)
[legion] inventoried 0 files for proberepo (2 stale rows pruned)
exit=0
```

That is the exact failure mode the missing-workdir guard (0bdfcb7) was added to prevent -- a transient mount/permission failure silently destroying derived state, with exit 0. The guard has a hole for root-level IO errors.

Suggested fix mirrors what #719 already does in its twin walk: a root-level walk error is distinguishable via `ignore::Error::depth() == Some(0)` (see `src/etc.rs` `failed_repos` handling). Make `walk_repo` return `Result` and fail on a depth-0 error, or at minimum refuse to prune when the walk errored and produced zero entries.

### LOW (non-blocking)

1. **symbol_count zeroing on failed SCIP runs.** Every walk entry carries `symbol_count: 0` and the upsert's `DO UPDATE SET symbol_count = excluded.symbol_count` overwrites unconditionally; the enrich pass runs only after successful indexing. A machine whose scip indexer is missing/broken zeroes every count on each failed run -- silently wrong (0 is a meaningful value: "no SCIP coverage") until the next successful index. Consider doing the reset inside the update pass instead (zero the rows NOT present in `counts`, only when `indexed > 0`): a failed SCIP run then degrades to stale counts rather than false zeros, and the self-heal for files that drop out of the blob is preserved.
2. **Redundant index.** `idx_file_inventory_repo` duplicates the prefix of the `PRIMARY KEY (repo, path)` autoindex; repo-only lookups are already served. `idx_file_inventory_repo_ext` is fine.
3. **Backslash mangling on unix.** `inventory.rs` `.replace('\\', "/")` turns a unix filename that legitimately contains a backslash into a phantom subdirectory path. Gate to `cfg(windows)` or build the rel path from `components()`.
4. **Reproducibility doc claim vs WalkBuilder defaults.** `parents` (reads .gitignore in directories ABOVE the repo root -- machine-layout dependent) and `git_exclude` (`.git/info/exclude` -- per-checkout local) both default ON; `git_global(false)` alone does not deliver the doc's "reproducible across machines" claim. Set `parents(false)` and decide `git_exclude` deliberately, or soften the doc.
5. **Prune is three statements, not one transaction.** Temp-table clear, insert-tx, then the DELETE each run separately; a crash between the insert commit and the DELETE leaves stale rows until the next index (self-heals, no corruption). Temp-table DDL is transactional in SQLite, so the whole prune could be one tx for free.

### Merge-collision heads-up (for whoever merges second)

#718 and #719 collide textually: `Cargo.toml` (both add `ignore = "0.4"` at the same spot), `src/main.rs` (adjacent `mod` lines), and `src/cli/index_cmd.rs` (this PR rewrites `handle_index`; #719 touches the same lines, including the old `langs.is_empty()` error arm). Resolution must keep this PR's restructured `handle_index`.

**Verdict: changes requested on the MED (posted as comment -- same-account formal review blocked); everything else non-blocking.**

ssilvius:
UPDATED PR BODY (supersedes the description above; the branch gained review-fix commits and a history rewrite for authorship repair -- see #723):

## Summary

Implements #705: a `file_inventory` table populated on every `legion index` run -- one row
per non-ignored file in the watched repo, pruned when files disappear. This is the always-
fresh structured substrate that `sym tree` (#706) and `sym etc find-file` (#709) will query.
New `src/inventory.rs` (gitignore-aware walk + ext-to-lang mapping) and `src/db/inventory.rs`
(DDL + upsert/list/prune), wired into `handle_index` after -- and independent of -- the SCIP
loop, which docs-only repos now skip instead of erroring.

## Acceptance criteria mapping

### 1. Inventory populated for a real repo on legion index, one row per non-ignored file

`inventory::walk_repo` drives an `ignore::WalkBuilder` over the workdir (require_git(false),
hidden(false), explicit `.git` filter_entry) and returns one FileInventoryEntry per file;
`handle_index` upserts the batch in one transaction and logs the count. Dotfiles like
`.github/workflows/ci.yml` are included -- the pre-commit review caught that WalkBuilder's
hidden(true) default silently excluded them, fixed with a pinning test (dotfiles_are_included).
Evidence: tests walk_produces_one_row_per_file, dotfiles_are_included,
gitignored_files_are_excluded (src/inventory.rs).

### 2. Non-SCIP files present with lang = None, per-file lang from ext mapping

Per-file `lang` comes from the new `lang_for_ext` (src/inventory.rs:22), a flat ext-to-SCIP-tag
map -- NOT from `detect_languages`, which is repo-level marker sniffing and cannot classify
individual files. `.sh`/`.md`/`.toml`/`.yaml`/`.css` all return None and land as inventory rows
with lang NULL.
Evidence: tests sh_and_md_have_no_lang, lang_for_ext_unknown_returns_none.

### 3. Docs-only repo gets a populated inventory instead of an error

`handle_index` previously hard-errored when `detect_languages` returned empty; the guard is now
an eprintln + skip of the SCIP block only, and the inventory walk runs unconditionally BEFORE
it (reordered during review: the old placement let the all-indexers-failed hard error skip the
inventory, contradicting this PR's own independence claim). The "markers present but every
indexer failed" case still hard-errors, with its inventory already current. The exit-code
contract flip is pinned at the CLI boundary by a real integration test through the binary --
added after review correctly flagged that the walk-level unit test could not cover it.
Evidence: integration test index_docs_only_repo_succeeds_and_inventories
(tests/integration/index_telemetry.rs); unit test docs_only_repo_produces_inventory
(walk level).

### 4. Re-index is idempotent; deleted files are pruned

Upsert is keyed ON CONFLICT(repo, path) DO UPDATE (one prepared statement, one transaction),
deliberately NOT overwriting symbol_count -- that column is owned by the enrich pass
(`update_file_symbol_counts`, reset+apply in one transaction) so a failed SCIP run cannot zero
prior enrichment and a deleted-last-definition file cannot keep a stale count. After each walk,
`prune_file_inventory` deletes rows absent from the live set -- via a temp table to dodge
SQLite's bound-variable limit -- and the prune is GATED on a clean walk: `walk_repo` returns
`WalkOutcome{entries, walk_errors}` and a partial walk (unreadable root, transient IO) refuses
eviction instead of wiping the table with exit 0, the data-loss bug smugglr's review reproduced
empirically. Two review discoveries are pinned by regression tests: the ignore crate does NOT
skip `.git/` once hidden(false) is set (git_dir_is_excluded_even_with_hidden_false), and the
unreadable-workdir prune refusal (index_unreadable_workdir_skips_prune_instead_of_wiping_inventory).
Evidence: db tests in src/db/inventory.rs; the two named regression tests.

### 5. Tests pass; simplify + review done; PR linked

1180 unit tests pass plus the CLI integration tests, clippy -D warnings and fmt clean. Review
ran the full four-dimension + adversarial-refuter pipeline plus smugglr's Rust lane; every
surviving finding landed in-branch: inventory-before-SCIP ordering, docs-only CLI integration
test, walk-errors prune gating (smugglr's reproduced wipe bug), symbol_count column ownership
with parse-success gating, prepared-statement upsert, dropped redundant index, borrowed prune
paths, mtime doc accuracy, symlink pinning test, strengthened mutation-visibility test.
Follow-ups filed rather than scope-crept: #721 (repo-wide busy_timeout), #722 (per-repo index
lock for the prune race -- land before/with #706 when the table gains readers), #723 (test
fixtures poisoning the real .git/config, discovered live this session).

## Not done

- `symbol_count` population from SCIP blobs -- deferred to E6/E7 per the issue's Out of Scope;
  every row lands with symbol_count = 0.
- The read API (`list_file_inventory` / `InventoryFilter`) ships #[allow(dead_code)]-gated:
  its consumers are #706 (sym tree) and #709 (find-file), which remove the gate.
- No incremental single-file inventory update on `legion index --file` -- the full walk is
  cheap (one readdir pass) and runs on the existing repo-level path; per-file incrementalism
  can follow if telemetry shows index latency pain.

Closes #705.

PR #764, opened 2026-07-09
fix(#679): unify list/release liveness for persona wake leases

### LINKED ISSUE #679: leases list shows rows the release path considers not-live (display vs state mismatch)
## Summary

Split from #677. The wedge reaper in #677 releases a reaped session's persona
wake-lease correctly, but a separate observability inconsistency remains in the
lease CLI surface, observed during the #676 investigation:

- `legion watch leases list` shows lease rows with a uniform EXPIRES that
  ADVANCES on each invocation (e.g. 21:39 -> 21:45 across all personas), and
  keeps showing a row,
- while `legion watch leases release --persona <p> --signal <id>` reports
  "no live lease found for <p>/<id> (may already be released)" for a row the
  list still displays.

So the list view and the release path disagree about what counts as a "live"
lease. The list appears to render historical/all rows with a heartbeat-refreshed
expiry rather than the actual acquirable-lease state, which makes the CLI
misleading for diagnosing a stuck persona.

## Acceptance criteria

- [ ] `leases list` reflects actual live-lease state (filter expired/released, or
      clearly mark non-live rows) so it agrees with the release path
- [ ] `leases list` and `leases release` share one definition of "live"
- [ ] A released or expired lease does not appear as live in the list
- [ ] Test covering the list/release agreement

ssilvius:
## Summary

Persona wake leases had two independent liveness checks that could disagree: `list_persona_leases`
tested both `deleted_at IS NULL` and `expires_at > now`, while the three release paths
(`release_persona_lease`, `release_persona_lease_if_owner`, `release_persona_leases_by_host`) tested
only `deleted_at IS NULL`. An expired-but-undeleted lease was already invisible to `list` yet still
reported as "releasable" (`true`) by `release_persona_lease` -- a live-state drift between what the
CLI displays and what the release paths act on. This fix introduces one named constant,
`LIVE_LEASE_WHERE` (`src/db/wake.rs:55`), holding the single liveness predicate, and every one of the
five call sites now shares it via `format!` interpolation, so there is exactly one definition of "this
lease is live" instead of five near-duplicate copies that can drift.

### 1. `leases list` reflects actual live-lease state (filter expired/released)

`list_persona_leases` already filtered on `deleted_at IS NULL AND expires_at > ?1` before this change
(that half of the bug was never in `list`); what changed is that it now draws that filter from the
shared `LIVE_LEASE_WHERE` constant (`src/db/wake.rs:133`, `:151`) instead of an inline literal, so the
list's notion of "live" cannot silently drift from the release paths' notion in a future edit. Test
`persona_lease_list_omits_expired` (pre-existing) and the new
`persona_lease_never_renders_as_live_after_release_or_expiry` (`src/db/wake.rs:212`) both assert
`list_persona_leases` returns no row for an expired-or-released lease, i.e. it reflects actual
live-lease state rather than raw table contents with a heartbeat-refreshed display expiry.

### 2. `leases list` and `leases release` share one definition of "live"

`LIVE_LEASE_WHERE` (`src/db/wake.rs:55`, `"deleted_at IS NULL AND expires_at > ?"`) is now the single
predicate shared by `list_persona_leases` (`src/db/wake.rs:133`, `:151`), `release_persona_lease`
(`src/db/wake.rs:50`), `release_persona_lease_if_owner` (`src/db/wake.rs:81`), and
`release_persona_leases_by_host` (`src/db/wake.rs:106`). Before, only `list_persona_leases` checked
`expires_at`; the three release paths checked only `deleted_at`, so an expired-but-undeleted row was
"live" to release and "gone" to list. All five queries now use the identical predicate string, so
there is exactly one definition of live, not five copies that can diverge.

### 3. A released or expired lease does not appear as live in the list

`persona_lease_never_renders_as_live_after_release_or_expiry` (`src/db/wake.rs:212`) is the direct
test for this criterion: it drives one lease through explicit release and a second through TTL expiry
with no explicit release, then asserts `list_persona_leases` shows neither as live in either case --
covering both routes by which a lease can stop being live, not just one.

### 4. Test covering the list/release agreement

`persona_lease_release_and_list_agree_on_expired_lease` (`src/db/wake.rs:162`) and
`persona_lease_release_if_owner_and_list_agree_on_expired_lease` (`src/db/wake.rs:192`) each acquire a
zero-TTL lease, let it expire, and assert both `list_persona_leases` and the corresponding release
function (unscoped and host-scoped respectively) agree it is gone: `list` already shows it empty, and
release now reports `false` ("no live lease found") instead of the pre-fix `true`. `cargo test
wake::` (`src/db/wake.rs`) runs all three new tests plus the 34 pre-existing wake tests green.

## Deliberately not done

- Did not touch `release_persona_leases_by_host` wiring -- it remains `#[allow(dead_code)]`, unused
  until a future SIGTERM handler calls it (pre-existing, out of scope for #679); only its liveness
  predicate was unified.
- Did not add a grace period or change TTL semantics -- the fix is purely about the two liveness
  checks (`deleted_at`, `expires_at`) agreeing across all five call sites, not about changing what
  "expired" means.
- Did not refactor `map_persona_lease_row` or the `PersonaWakeLease` struct; unrelated to the liveness
  drift.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test wake::` (37/37
passing, including the 3 new tests) all run clean on HEAD `82e79692db6279ffa6bf2281aea6db698a1ea9ee`.

Closes #679

### DIFF
--- a/src/db/wake.rs
+++ b/src/db/wake.rs
@@ -39,6 +39,21 @@ fn map_persona_lease_row(row: &rusqlite::Row<'_>) -> rusqlite::Result<PersonaWak
     })
 }
 
+/// The single definition of "this lease is live": not soft-deleted and not
+/// past its TTL. Shared verbatim by every reader/writer that needs to agree
+/// on liveness -- `list_persona_leases` (what the CLI displays),
+/// `release_persona_lease` / `release_persona_lease_if_owner` (what counts as
+/// releasable), and `release_persona_leases_by_host` -- so a released-or-
+/// expired lease can never render as live in one code path while another
+/// still treats it as gone (#679: the list and release paths previously
+/// diverged because release only checked `deleted_at`, ignoring `expires_at`,
+/// so an expired-but-undeleted row was invisible to `list` yet still
+/// "releasable" to `release`).
+///
+/// Uses one unnumbered `?` placeholder -- callers bind `now` at the position
+/// where this fragment lands in the finished SQL text.
+const LIVE_LEASE_WHERE: &str = "deleted_at IS NULL AND expires_at > ?";
+
 /// `persona_wake_leases` (#308) and `wake_attempts` (#487) tables.
 pub(super) fn create_tables(conn: &Connection) -> Result<()> {
     // Migration 17: Persona wake leases for cluster-wide wake coordination (#308).
@@ -205,7 +220,10 @@ impl Database {
     }
 
     /// Soft-delete one lease by (persona_id, signal_id). Returns true if a
-    /// matching live lease existed. Idempotent on an already-released lease.
+    /// matching *live* lease existed (per [`LIVE_LEASE_WHERE`] -- the same
+    /// predicate `list_persona_leases` uses), false when it was already
+    /// released or had already expired. Idempotent on an already-released
+    /// lease.
     ///
     /// Unscoped by host -- used by the operator CLI to forcibly drop any
     /// stuck lease. The watch reaper uses `release_persona_lease_if_owner`
@@ -213,11 +231,14 @@ impl Database {
     /// accidentally release the winner's row.
     pub fn release_persona_lease(&self, persona_id: &str, signal_id: &str) -> Result<bool> {
         let now = Utc::now().to_rfc3339();
-        let updated = self.conn.execute(
+        let sql = format!(
             "UPDATE persona_wake_leases \
-             SET deleted_at = ?1, updated_at = ?1 \
-             WHERE persona_id = ?2 AND signal_id = ?3 AND deleted_at IS NULL",
-            rusqlite::params![&now, persona_id, signal_id],
+             SET deleted_at = ?, updated_at = ? \
+             WHERE persona_id = ? AND signal_id = ? AND {LIVE_LEASE_WHERE}"
+        );
+        let updated = self.conn.execute(
+            &sql,
+            rusqlite::params![&now, &now, persona_id, signal_id, &now],
         )?;
         Ok(updated > 0)
     }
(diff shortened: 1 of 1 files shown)


ssilvius (commented):
Follow-up commit 8eccf9f fixes a MED regression the review confirmed on the previous head (82e7969): unifying LIVE_LEASE_WHERE across all five call sites had gated release_persona_lease_if_owner's WRITE on expires_at too. That function is the reaper's own-row finalization write, not a liveness read -- gating it meant an already-expired-but-owned lease no longer got soft-deleted on reap, and the host-blind heartbeat_persona_leases (deleted_at IS NULL only) then bumped expires_at forward on the same tick_health cycle, resurrecting it into a permanent ghost in `leases list`.

Fix: release_persona_lease_if_owner no longer uses LIVE_LEASE_WHERE -- it finalizes on acquired_by_host + deleted_at IS NULL alone, unconditional on expiry. list_persona_leases, release_persona_lease (operator CLI), and release_persona_leases_by_host (shutdown) are unchanged and still share LIVE_LEASE_WHERE, so the PR body's criterion #2 mapping is now stale for release_persona_lease_if_owner specifically -- noting it here rather than rewriting the body.

Added persona_lease_reap_then_heartbeat_does_not_resurrect_expired_lease as the direct regression test (drives reap-then-heartbeat against a 0-TTL owned lease, asserts it never reappears in list_persona_leases), and flipped persona_lease_release_if_owner_and_list_agree_on_expired_lease -> persona_lease_release_if_owner_finalizes_an_already_expired_lease since the reaper path must NOT agree with list on an expired lease anymore.

Not fixed here, flagged as an out-of-scope residual: a lease that is never reaped at all (daemon restart losing tracked held_leases, or a crash before reap_finished runs) stays deleted_at IS NULL, and heartbeat_persona_leases -- host-scoped but liveness-blind -- can still keep such a row artificially fresh indefinitely. Closing that fully would mean scoping heartbeat itself (e.g. AND expires_at > now, or restricting to tracked sessions), which is a separate, narrower-blast-radius change than this one.

All tests green: cargo test -- 1383 unit + 276 integration passed, 0 failed. clippy -D warnings and fmt --check clean.

